PR #77, opened 2018-02-12
Created the Uber jar for compiled java code and dependencies using jar.exe

sachi-sharma:
Created the Uber jar using the same jar.exe
Path : root


mcoblenz:
Are you asking to merge in changes from master into the branch? In general, there's no need for me to review commits/merges into branches; only into master.

PR #81, opened 2018-02-23
Added test cases for compilation check

sachi-sharma:


src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -206,7 +205,9 @@ object Main {
             println(usage)
             sys.exit(0)
         }
-
+        compileProgram(args)
+    }
+    def compileProgram(args: Array[String]): Boolean = {
```
mcoblenz: If protoc returns nonzero, we should return false.

PR #92, opened 2018-05-30
Better error message when client specifies improper connection information

broomweed:
And also a bunch of library files got added by IntelliJ, somehow.


.idea/sbt.xml
```diff
@@ -10,9 +10,8 @@
             <option value="$PROJECT_DIR$/project" />
           </set>
         </option>
-        <option name="sbtVersion" value="1.0.3" />
+        <option name="sbtVersion" value="1.0.4" />
         <option name="useAutoImport" value="true" />
-        <option name="useOurOwnAutoImport" value="true" />
```
mcoblenz: Why the change here?

.idea/.name
```diff
@@ -1 +1 @@
-Obsidian
```
mcoblenz: Why is this change here?

.idea/libraries/sbt__com_google_protobuf_protobuf_java_3_5_1_jar.xml
```diff
@@ -0,0 +1,11 @@
+<component name="libraryTable">
```
mcoblenz: Actually I looked at this and I think we should ignore these files (gitignore) instead: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems

mcoblenz:
Also, this change should be split into two: one to fix the error message and one to fix the library files.

broomweed:
OK, I'm going to close this one and create two new branches instead.

PR #93, opened 2018-05-31
Issue 91

suzzglennon:
@mcoblenz Sorry for the delay, but here is the last of the code I hadn't pushed yet! Please let me know if there's anything else you need from me.


mcoblenz:
I've finished this off and merged it in with d74d8c57.

mcoblenz:
I suppose I should have done the remaining work on this branch rather than opening a new one, but that's probably why GitHub isn't aware that this has all been merged in.

PR #97, opened 2018-06-01
Issue 44

pkoronkevich:
checkTransactionArgShadowing checks whether the arguments of a given transaction match the fields of the possible states it could start in.

Example now gives the following error message:
At 8.2: The transaction parameter 'x' in 't' also declared in contract at line 4. Consider changing the transaction parameter.

Added the given example in issue as a test (ArgumentShadowing.obs). A test had to be changed since it had this shadowing issue. Test changed to no longer shadow arguments was MaybeDroppedResource.obs. A new test added was DroppedResourceArgShadowing.obs which contains the original test that had argument shadowing.


src/main/scala/edu/cmu/cs/obsidian/typecheck/Error.scala
```diff
@@ -29,6 +29,11 @@ case class NoMainContractError() extends Error {
 case class ShadowingError(fieldName: String, stateName: String, prevLine: Int) extends Error {
     val msg: String = s"Field '$fieldName' in '$stateName' also declared in contract at line $prevLine. Consider removing declaration in state."
 }
+
+case class ArgShadowingError(arg: String, transactionName: String, prevLine: Int) extends Error {
+    val msg: String = s"The transaction parameter '$arg' in '$transactionName' also declared in contract at line $prevLine. Consider changing the transaction parameter."
```
mcoblenz: Need to insert "is": "…is also declared…"

resources/tests/type_checker_tests/DroppedResourceArgShadowing.obs
```diff
@@ -0,0 +1,23 @@
+resource contract Money {}
```
mcoblenz: What is this test case for? Is there a relationship between dropping resources and shadowing? If not, let's skip this extra test case.

Issue #99, opened 2018-06-05
Need default constructor


pkoronkevich:
**Describe the bug**
The invokeConstructor() method doesn't take any arguments; but the constructor generated may take them.

**To Reproduce**
Any contract with a constructor that takes arguments, ex:

`main contract C {
    int a;
    int b;

    C(int x, int y) {
        a = x;
        b = y;
    }
}`

will result in an error in invokeConstructor() : 

`C.java:23: error: method new_C in class C cannot be applied to given types;
        this.new_C();
            ^
  required: BigInteger,BigInteger
  found: no arguments
  reason: actual and formal argument lists differ in length`

**Expected behavior**
A new_C() method generated that assigns default values to fields. So generated code should now have a default constructor.

PR #100, opened 2018-06-06
Generate default constructor for main contracts that don't have one

broomweed:
This fixes issue #99, by adding a default constructor to main contracts which don't have one already defined, and by not generating the `invokeConstructor()` method on non-main contracts, since they don't need it and can use the base-class implementation that does nothing.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -655,10 +655,17 @@ class CodeGen (val target: Target) {
 
         for (decl <- aContract.declarations) {
             translateDeclaration(decl, newClass, translationContext, aContract)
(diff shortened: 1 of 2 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -655,12 +655,19 @@ class CodeGen (val target: Target) {
 
         for (decl <- aContract.declarations) {
             translateDeclaration(decl, newClass, translationContext, aContract)
-            if (decl.isInstanceOf[Constructor] && (!generated)) {
+            if (decl.isInstanceOf[Constructor] && (!generated) && aContract.modifiers.contains(IsMain())) {
```
mcoblenz: Could you please add an isMain() convenience method to Contract? There's already an isResource() method to model it on.

PR #103, opened 2018-06-08
Issue 98

pkoronkevich:
To resolve this issue, I had to refactor several things.

In order to properly deserialize the generated code `byte[][] args` in `init` , I needed the types of the arguments to the main contract constructor. To ensure the proper types, the type checker must now enforce that main contracts have only one constructor. Thus, `byte[][] args` should always have the same number of arguments with the same types.

To do this, I added a line in `checkContract` in Checker.scala to check if a main contract had more than one constructor - if so, logged a new `MultipleConstructorsError`. Also added a new typechecker test.

Next, there were two methods in CodeGen.scala, `translateMainConstructor` and `translateConstructor`, which were very similar, so I merged them into a single method `translateConstructor` that now takes a contract to check if it is the main contract. If it is, it adds a private `new_ContractName` method which is used when the main contract is deployed.

Finally, `generateInitMethod` in CodeGen.scala changed in order to deserialize arguments properly. It has to take the types of the arguments, which is determined in `generateMainServerClassMethods` since this is the only call site of `generateInitMethod` . The inherited method from ChaincodeBaseMock.java had to be updated to throw a Protocol Buffer exception. Then I added a check in the generated code that `byte[][] args` has the right length before indexing and deserializing elements of args according to their type.

I also added a compiler test to test this particular case.

### DIFF
--- a/resources/tests/type_checker_tests/MultipleConstructors.obs
+++ b/resources/tests/type_checker_tests/MultipleConstructors.obs
@@ -1,0 +1,14 @@
+main contract C {
+    int x;
+    int y;
+
+    C(int first) {
+        x = first;
+        y = 0;
+    }
+
+    C(int first, int second) {
+        x = first;
+        y = second;
+    }
+}
--- a/resources/tests/type_checker_tests/MultipleConstructors.obs
+++ b/resources/tests/type_checker_tests/MultipleConstructors.obs
@@ -1,3 +1,5 @@
+// testing for main contracts to only have one constructor
+
 main contract C {
     int x;
     int y;
@@ -6,7 +8,7 @@ main contract C {
         x = first;
         y = 0;
     }
-
+    // error: main contracts should have one constructor
     C(int first, int second) {
         x = first;
         y = second;
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -1633,31 +1633,8 @@ class CodeGen (val target: Target) {
 
     /* the local context at the beginning of the method */
 
-    /* The java constructor in the main contract runs every transaction.
-     * The obsidian constructor only runs when the contract is deployed.
+    /* The obsidian constructor only runs when the contract is deployed.
      * Thus, the obsidian constructor must be placed in a distinct method. */
-    private def translateMainConstructor(
-                                            c: Constructor,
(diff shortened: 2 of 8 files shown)


resources/tests/type_checker_tests/MultipleConstructors.obs
```diff
@@ -0,0 +1,14 @@
+main contract C {
```
mcoblenz: Add comment indicating what error is expected here (and what this test case is testing).

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -732,8 +732,22 @@ class CodeGen (val target: Target) {
         /* run method */
         generateRunMethod(newClass, translationContext, stubType)
 
+        // need to gather the types of the main contract constructor in order to correctly deserialize arguments
+        var constructorTypes: List[ObsidianType] = List.empty
+        for (decl <- translationContext.contract.declarations) {
+          decl match {
+            case const: Constructor => {
+              for (arg <- const.args) {
```
mcoblenz: I suggest using map here instead of iterating over the args.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -732,8 +732,22 @@ class CodeGen (val target: Target) {
         /* run method */
         generateRunMethod(newClass, translationContext, stubType)
 
+        // need to gather the types of the main contract constructor in order to correctly deserialize arguments
+        var constructorTypes: List[ObsidianType] = List.empty
+        for (decl <- translationContext.contract.declarations) {
```
mcoblenz: This outer loop is kind of weird because you want to assume there is only one constructor, but the code actually corresponds with concatenating arguments from ALL constructors, which doesn't make any sense. How about stopping after you find the first one?

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -732,8 +732,22 @@ class CodeGen (val target: Target) {
...
+              }
+            }
+            case _ => ()
+          }
+
+        }
+
         /* init method */
-        generateInitMethod(newClass, stubType)
+        generateInitMethod(newClass, stubType, constructorTypes.reverse)
```
mcoblenz: Using map will obviate the need to call reverse.

PR #109, opened 2018-06-12
Issue 104

pkoronkevich:
I changed `translateTransDecl` to no longer check if transactions existed in only one possible state, which then it would call `translateTransDeclInPossibleState`. This method generated the transaction in an inner state class. This caused a "method not found" error in the `run` method generated in the top level contract class. I instead transferred the contents of `translateTransDeclInPossibleState` to `translateTransDecl`, and inserted a run time check for transactions only available in certain states.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -1709,10 +1709,10 @@ class CodeGen (val target: Target) {
 
         availableIn match {
             case Some(states) => {
-                var cond: IJExpression = JExpr.FALSE
+                var cond: IJExpression = JExpr.TRUE
                 // check if the current state is in any of the possible states
                 for (st <- states) {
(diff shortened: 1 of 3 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -1690,19 +1690,36 @@ class CodeGen (val target: Target) {
...
+
+        // Dynamically check the state
+        val availableIn: Option[Set[Identifier]] = tx.availableIn
+
+        availableIn match {
+            case Some(states) => {
+                var cond: IJExpression = JExpr.FALSE
+                // check if the current state is in any of the possible states
+                for (st <- states) {
+                    cond = JOp.cor(JExpr.invoke(getStateMeth).ne(translationContext.getEnum(st._1)), cond)
```
mcoblenz: You want and here, not or. Here is the generated code:

````
if ((getState()!= State_A.S2)||(getState()!= State_A.S1)) {
            throw new edu.cmu.cs.obsidian.chaincode.BadTransactionException();
        }
````

PR #112, opened 2018-06-12
Issue 101

broomweed:
Added `InterfaceContractType` cases to several pattern-matches. Often, it just fails if encountering an interface type, since we don't know how to serialize/deserialize an interface, for instance. But we might add support for interface serialization in the future.

I also added `OwnershipTransfer` to the AST pattern-match block, which asserts false since the syntax for this is going to change.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -1573,6 +1573,7 @@ class CodeGen (val target: Target) {
                 addArgs(JExpr._new(model.ref(name)), args, translationContext, localContext)
             case Parent() => assert(false, "Parents should not exist in code generation"); JExpr._null()
             case Disown(e) => recurse(e)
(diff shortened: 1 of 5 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -1191,7 +1191,7 @@ class CodeGen (val target: Target) {
 
         def handleNonPrimitive(name: String, n: ObsidianType): Unit = {
             // foo = new Foo(); foo.initFromArchive(archive.getFoo());
-            val javaFieldTypeName = javaFieldType.fullName()
+            // val javaFieldTypeName = javaFieldType.fullName()
```
mcoblenz: Please delete this line and the commented-out line above it.

broomweed: I think the commented-out line above it is to say what code will be generated from jCodeModel. But, yeah, I'll delete the other one -- it was flagged as an unused declaration.

PR #118, opened 2018-06-18
Issue 19

pkoronkevich:
In order to support imports in all contracts, I added a stage to code generation for processing imports. At multiple points in code generation, `parseFileAtPath` was called on the imports in the program; however, it is much simpler to parse imports once and add them to the program.

I did this by adding an import processing stage right after parsing and before AST transformation. This allowed me to remove duplicate code from AST transformation, code generation, and protobuf generation, and instead add asserts that program imports are empty after processing. 

In order to properly generate stub contracts, however, I needed to add an `IsImport()` tag to contracts to know which contracts needed stub contracts generated.

ImportProcessor generates a sequence of contracts for each import, keeping track of imports already seen as well as contract names already seen to avoid cyclic imports and name shadowing issues. It recursively gathers contracts from imports from inside other imports as well, and adds the IsImport tag to all of them. 

There was an issue that arose in code generation that there must only be one main contract for building clients, so it removes any IsMain tag from imported contracts.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -708,12 +708,12 @@ class CodeGen (val target: Target) {
 
         // need to gather the types of the main contract constructor in order to correctly deserialize arguments
         // find the first declaration that is a constructor
-        val constructor: Option[Declaration] = translationContext.contract.declarations.find(decl => decl.isInstanceOf[Constructor])
+        val constructor: Option[Declaration] = translationContext.contract.declarations.find(_.isInstanceOf[Constructor])
         // gather the types of its arguments
         val constructorTypes: Seq[ObsidianType] =
             constructor match {
-                case Some(constr: Constructor) => constr.args.map(a => a.typ)
-                case None => List.empty
+                case Some(constr: Constructor) => constr.args.map(_.typ)
+                case _ => List.empty
             }
 
         /* init method */
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -98,7 +98,7 @@ class CodeGen (val target: Target) {
         }
 
         for (c <- program.contracts) {
-            // TODO : generate code for interfaces
+            // TODO : generate code for interfaces (issue #117)
             if(!c.isInterface) {
                 translateOuterContract(c, programPackage, protobufOuterClassName, contractNameResolutionMap, protobufOuterClassNames)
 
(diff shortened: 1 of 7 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -88,20 +88,24 @@ class CodeGen (val target: Target) {
...
         }
 
-        for (imp <- program.imports) {
-            if(!imp.name.contains("/java-utilities/"))
-                translateImport(programPackage, imp, contractNameResolutionMap, protobufOuterClassNames)
-        }
-
         for (c <- program.contracts) {
-            translateOuterContract(c, programPackage, protobufOuterClassName, contractNameResolutionMap, protobufOuterClassNames)
+            // TODO : generate code for interfaces
```
mcoblenz: Please open a Github issue and reference it here.

src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala
```diff
@@ -0,0 +1,62 @@
...
+            contracts match {
+                case Left(msg) => throw new ImportException(msg)
+                case Right(cs) => allContracts = cs ++ allContracts
+            }
+        })
+
+        Program(Seq.empty, allContracts)
+    }
+
+    def processImport(imp: String, seen: Seq[Import], contractNames: Seq[String]) : Either[String, Seq[Contract]] = {
```
mcoblenz: Use a more meaningful name than "imp". Perhaps "importPath"?

src/main/scala/edu/cmu/cs/obsidian/parser/SymbolTable.scala
```diff
@@ -91,7 +91,7 @@ class StateTable(
 
 class ContractTable(
         val contract: Contract,
-        symbolTable: SymbolTable,
+        var symbolTable: SymbolTable,
```
mcoblenz: Why this change?

src/main/scala/edu/cmu/cs/obsidian/protobuf/ProtobufGen.scala
```diff
@@ -14,33 +14,15 @@ class Unimplemented extends Exception {}
...
-
-            val messages = ast.contracts.map(translateContract)
-            (new Protobuf(messages), filename)
-        })
+        assert(program.imports.isEmpty, "Imports should be empty after processing.")
 
         val messages = program.contracts.map(translateContract)
 
-        val result = protobufs :+ ((new Protobuf(messages), sourceFilename))
+        val result: Seq[(Protobuf, String)] = (new Protobuf(messages), sourceFilename) +: Seq.empty
```
mcoblenz: I suggest `Seq((new Protobuf(messages), sourceFilename))` rather than an append just to make a sequence.

PR #119, opened 2018-06-21
Unified state union types with single-state types; now all state types can represent unions of states.

mcoblenz:
This is in preparation for re-aligning the type system to the current language design.

I thought initially that the code generator would need changes, but it appears that is not the case.


mcoblenz:
Merged via command line. Not sure why the GitHub interface doesn't know that.

PR #122, opened 2018-06-21
Generate real Hyperledger chaincode

broomweed:
Add a compiler option `--hyperledger` which causes the compiler to produce actual Hyperledger-Fabric-compatible Java chaincode. At the moment, the chaincode can't do anything besides be initialized, and will return an empty `byte[]` array for all requests. But, it's a start!

Note that due to the large number of dependencies for the Hyperledger Fabric library, the chaincode needs to be compiled with Gradle, using the `buildscript/build.gradle`, which will handle downloading all the proper dependencies and compiling them into a JAR.

There is also a shell script, `buildscript/startfabric.sh` which handles starting a Fabric network in dev-mode and installing and initializing the chaincode. (Note that the script should only be used for testing purposes, as it wipes out your entire `/var/hyperledger/production` directory every time in order to reinstall the chaincode under the same name.)


mcoblenz:
The compiler appears to complete successfully, but I'm not getting a build/libs/chaincode.jar file. The output is as follows:

Generate mock chaincode: false
edu/cmu/cs/obsidian/generated_code/IntContainer.java
> Task :compileJava UP-TO-DATE
> Task :processResources NO-SOURCE
> Task :classes UP-TO-DATE
> Task :jar UP-TO-DATE
> Task :startScripts UP-TO-DATE
> Task :distTar UP-TO-DATE
> Task :distZip UP-TO-DATE
> Task :assemble UP-TO-DATE
> Task :compileTestJava NO-SOURCE
> Task :processTestResources NO-SOURCE
> Task :testClasses UP-TO-DATE
> Task :test NO-SOURCE
> Task :check UP-TO-DATE
> Task :build UP-TO-DATE

BUILD SUCCESSFUL in 1s
5 actionable tasks: 5 up-to-date
gradle exited with value 0

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -730,8 +734,21 @@ class CodeGen (val target: Target) {
     }
 
     private def generateMainServerClassMethods(newClass: JDefinedClass, translationContext: TranslationContext): Unit = {
-        newClass._extends(model.directClass("edu.cmu.cs.obsidian.chaincode.ChaincodeBaseMock"))
-        val stubType = model.directClass("edu.cmu.cs.obsidian.chaincode.ChaincodeStubMock")
+
+        System.out.print("Generate mock chaincode: ")
```
mcoblenz: "mock chaincode: false" is a little surprising when in Hyperledger mode. How about saying what it IS doing instead of what it is NOT doing?

PR #139, opened 2018-07-02
Add @ notation to constructors.

pkoronkevich:
Add @ notation to constructors, i.e. `owned C() {...}` becomes `C@Owned() {...}` and `C() ends in S {...}` becomes `C@S() {...}`. Refactored some things in the checker as well - `InvokableDeclarations` no longer have an `endsInState` field; instead, the type is constructed while parsing.

This eliminates the need for `simpleOf` and `simpleOfWithStateNames` in the type checker, which were helpers to construct these types. These functions were basically copied to `parseTransDecl` to create fields `thisType` (which is the type the transaction is available in) and `thisFinalType` (which is the type the transaction ends in). 

There was already a mechanism to extract a type based on the permission within parsing, so I used that for extracting the type of the constructor.

Since the syntax changed, some of the tests had to change. Currently, we don't support state unions in the parser, so the `EndsInStateUnion.obs` test will have to change back once this is supported.

### DIFF
--- a/resources/tests/type_checker_tests/EndsInStateUnion.obs
+++ b/resources/tests/type_checker_tests/EndsInStateUnion.obs
@@ -1,6 +1,6 @@
 main contract C1 {
 
-    // error: constructor for 'C1' ends in S2, S3 instead of S1, S2
+    // error: constructor for 'C1' ends in S2, S3 instead of S1
     C1@S1(bool b){
         if(b) {
             ->S2;
--- a/resources/tests/type_checker_tests/EndsInStateUnion.obs
+++ b/resources/tests/type_checker_tests/EndsInStateUnion.obs
@@ -1,7 +1,7 @@
 main contract C1 {
 
     // error: constructor for 'C1' ends in S2, S3 instead of S1, S2
-    C1(bool b) ends in S1, S2 {
+    C1@S1(bool b){
         if(b) {
             ->S2;
         } else {
@@ -30,7 +30,7 @@ main contract C1 {
 contract C2 {
 
     // error: C2 ends in S1, S2 but must end in S1
-    C2(bool b) ends in S1 {
+    C2@S1(bool b) {
         if(b) {
             ->S1;
         } else {
--- a/src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
@@ -3,7 +3,7 @@ package edu.cmu.cs.obsidian.parser
 import scala.util.parsing.input.{NoPosition, Position}
 import edu.cmu.cs.obsidian.lexer.Token
 import edu.cmu.cs.obsidian.parser.Parser.{EndsInState, Identifier}
-import edu.cmu.cs.obsidian.typecheck.{ContractReferenceType, ContractType, StateType, InterfaceContractType, NonPrimitiveType, ObsidianType, Permission}
+import edu.cmu.cs.obsidian.typecheck._
 
 trait HasLocation {
     var loc: Position = NoPosition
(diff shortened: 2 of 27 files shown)


resources/tests/type_checker_tests/EndsInStateUnion.obs
```diff
@@ -1,7 +1,7 @@
 main contract C1 {
 
     // error: constructor for 'C1' ends in S2, S3 instead of S1, S2
```
mcoblenz: Update comment here.

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
```diff
@@ -2,8 +2,8 @@ package edu.cmu.cs.obsidian.parser
 
 import scala.util.parsing.input.{NoPosition, Position}
 import edu.cmu.cs.obsidian.lexer.Token
-import edu.cmu.cs.obsidian.parser.Parser.Identifier
-import edu.cmu.cs.obsidian.typecheck.{ObsidianType, Permission}
+import edu.cmu.cs.obsidian.parser.Parser.{EndsInState, Identifier}
+import edu.cmu.cs.obsidian.typecheck.{ContractReferenceType, ContractType, StateType, InterfaceContractType, NonPrimitiveType, ObsidianType, Permission}
```
mcoblenz: Perhaps use _ instead?

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
```diff
@@ -121,20 +119,33 @@ case class Func(name: String,
...
 }
 case class Transaction(name: String,
                        args: Seq[VariableDecl],
                        retType: Option[ObsidianType],
                        availableIn: Option[Set[Identifier]],
                        ensures: Seq[Ensures],
-                       endsInState: Option[Set[Identifier]],
                        body: Seq[Statement],
                        isStatic: Boolean,
+                       thisType: NonPrimitiveType,
```
mcoblenz: I think we can remove thisPermission and thisFinalPermission now, right?

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -1505,18 +1506,21 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
...
+                    stateTableOpt match {
+                        case None => logError(constr, StateUndefinedError(table.contract.name, stateName))
+                        case Some(_) => ()
+                    }
                 }
             }
+            case _ => ()
         }
 
         // TODO: make this owned?
```
mcoblenz: Can remove the TODO now.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -1529,6 +1533,24 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
...
+
+        if (constructors.length > 1 && contract.isMain) {
+            logError(contract, MultipleConstructorsError(contract.name))
+        }
+
+        val constructorsByArgTypes = constructors.groupBy(c => c.args.map(_.typ))
+        val matchingConstructors = constructorsByArgTypes.filter(_._2.size > 1)
+
+        matchingConstructors.foreach(typeAndConstructors => {
+            val highestLine = typeAndConstructors._2.maxBy(_.loc.line)
```
mcoblenz: Neat way of doing this. But "highest" is confusing because that might mean the line with lower line number. Consider "greatest" instead.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -1529,6 +1533,24 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
...
+    private def checkConstructors(constructors: Seq[Constructor], contract: Contract, table: ContractTable): Unit = {
+        if (constructors.isEmpty && table.stateLookup.nonEmpty) {
+            logError(contract, NoConstructorError(contract.name))
+        }
+
+        if (constructors.length > 1 && contract.isMain) {
+            logError(contract, MultipleConstructorsError(contract.name))
+        }
+
+        val constructorsByArgTypes = constructors.groupBy(c => c.args.map(_.typ))
```
mcoblenz: What if the types are not identical, but one is a subtype of the other? e.g. Foo@Owned, Foo@Unowned?

Issue #137, opened 2018-06-29
Support state union syntax


pkoronkevich:
Currently the parser doesn't support state union syntax, i.e. for constructors:

`C@(S1|S2)() {...}`

EndsInStateUnion test needs to be updated from `C1@S1` to `C1@(S1|S2)` once this change is complete.

mcoblenz:
Addressed with PR #145.

Issue #102, opened 2018-06-08
Add support for static assertions


mcoblenz:
Add support for static assertions, e.g. [x@Owned] checks to make sure x is an owning reference.

mcoblenz:
Syntax: 1 hour
Typechecker: 4 hours

Let's call it a day.

PR #145, opened 2018-07-06
Implemented static assertions

mcoblenz:
Addresses  #102. Also adds support for specifying sets of states when declaring types.


src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -502,7 +531,7 @@ object Parser extends Parsers {
 
     // maybe we can check here that the constructor has the appropriate name?
     private def parseConstructor = {
-        parseId ~ opt(AtT() ~! parseId) ~! LParenT() ~! parseArgDefList ~! RParenT() ~! LBraceT() ~! parseBody ~! RBraceT() ^^ {
+        parseId ~ opt(AtT() ~! parseIdAlternatives) ~! LParenT() ~! parseArgDefList ~! RParenT() ~! LBraceT() ~! parseBody ~! RBraceT() ^^ {
             case name ~ permission ~ _ ~ args ~ _ ~ _ ~ body ~ _ =>
```
pkoronkevich: this closes #137, so the `EndsInStateUnion.obs` test needs to be updated.

mcoblenz: Good catch, thanks. I'll update this test.

PR #149, opened 2018-07-06
Give a better error message when trying to assign a field in class body

broomweed:
I'm not 100% sure that this is the best way to accomplish this, but it wouldn't let me put the `failure` inside the `^^ { }` function, and this way does seem to work. Do let me know if there's a more idiomatic-Scala-parser-library way to do it, though.


mcoblenz:
Neat approach. I don't know the "right" way to do this, but if this works, that's good enough.

mcoblenz:
I'll add that another way would be to add support for the in the AST but then reject it at typechecking time, but that's a little more convoluted and probably worse overall.

PR #153, opened 2018-07-09
Issue 78

pkoronkevich:
To support typestate specification to arguments, I added a new `VariableDeclWithSpec` ast to have the type coming in and type going out of a arguments to a transaction. There is an optional argument "this" that can be the first argument to a transaction to replace `available in` and `ends in`.

### DIFF
--- a/resources/tests/parser_tests/InvokableSpec.obs
+++ b/resources/tests/parser_tests/InvokableSpec.obs
@@ -1,4 +1,3 @@
-/* for now, the ordering is 'requires -> returns -> ensures' */
 
 // NOTE: this file should parse, but not type-check
 
--- a/resources/tests/parser_tests/InvokableSpec.obs
+++ b/resources/tests/parser_tests/InvokableSpec.obs
@@ -11,8 +11,7 @@ contract C {
 
     C@S1() { return; }
 
-    // don't support state unions yet
-    C@S1() { return; }
+    C@(S1|S2)() { return; }
 
 
     transaction t() { return; }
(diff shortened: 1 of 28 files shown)


resources/tests/parser_tests/InvokableSpec.obs
```diff
@@ -11,8 +11,7 @@ contract C {
```
mcoblenz: Remove "for now…" comment at top of file

resources/tests/type_checker_tests/MaybeDroppedResource.obs
```diff
@@ -12,12 +12,12 @@ main resource contract Wallet {
 	 state Empty {
 	 }
 
-	 transaction loseMoney() available in Full, Empty {
+	 transaction loseMoney(Wallet@(Full|Empty) this) {
 	       // Error -- maybe losing Money!
```
mcoblenz: This is actually different and the compiler should now give an additional error: without the >> clause, the compiler should enforce that the state cannot change within the transaction.

src/main/scala/edu/cmu/cs/obsidian/typecheck/AstTransformer.scala
```diff
@@ -286,6 +286,10 @@ object AstTransformer {
                 val newTypChoice = transformType(table, lexicallyInsideOf, context, typ, s.loc)
                 val (newTyp, errors) = transformType(table, lexicallyInsideOf, context, typ, s.loc)
                 (oldDecl.copy(typ = newTyp).setLoc(oldDecl), context.updated(varName, newTyp), errors)
+            case oldDecl@VariableDeclWithSpec(typIn, typOut, varName) =>
+                val (newTypIn, errorsIn) = transformType(table, lexicallyInsideOf, context, typIn, s.loc)
+                val (newTypOut, errorsOut) = transformType(table, lexicallyInsideOf, context, typOut, s.loc)
+                (oldDecl.copy(typIn = newTypIn, typOut = newTypOut).setLoc(oldDecl), context.updated(varName, newTypIn), errorsIn ++ errorsOut)
```
mcoblenz: This line is pretty hard to read.

PR #155, opened 2018-07-10
Don't generate -a file.

### LINKED ISSUE #150: Avoid generating "-a" file when running on Hyperledger
The file is currently erroneously generated; it shouldn't exist.

broomweed:
Apparently I forgot to make a pull request for this.

`-a` is now called `chaincode_archive`, which is clearer. But we'll probably remove this eventually as well once we actually get some sort of serialization to the blockchain.

Fixes #150 .

### DIFF
--- a/Obsidian Runtime/src/Runtime/edu/cmu/cs/obsidian/chaincode/HyperledgerChaincodeBase.java
+++ b/Obsidian Runtime/src/Runtime/edu/cmu/cs/obsidian/chaincode/HyperledgerChaincodeBase.java
@@ -102,9 +102,8 @@ public abstract class HyperledgerChaincodeBase extends ChaincodeBase {
              * by Hyperledger. Because arguments are an array, we have to copy
              * all the remaining arguments into a new array. */
             String[] less_args = new String[args.length - 1];
-            for (int i = 0; i < args.length - 1; i++) {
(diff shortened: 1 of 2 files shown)


Obsidian Runtime/src/Runtime/edu/cmu/cs/obsidian/chaincode/HyperledgerChaincodeBase.java
```diff
@@ -98,7 +98,14 @@ public void delegatedMain(String args[]) {
 
         /* spin up server */
         try {
-            start(args);
+            /* Don't pass the first argument (archive path) to be processed
+             * by Hyperledger. Because arguments are an array, we have to copy
+             * all the remaining arguments into a new array. */
+            String[] less_args = new String[args.length - 1];
```
mcoblenz: Use System.arraycopy instead?

broomweed:
What was happening before was that the `-a $ip:7052 -i $CCNAME:$CCVERSION` was getting passed along at the `start(args)` line; the `start` function is part of the Hyperledger class and parses the arguments in its own way so it can find and connect to the peer server.

This version doesn't parse any other arguments after the first one, since it just relies on the Hyperledger part to handle the connection to the server rather than handling it itself.

PR #160, opened 2018-07-12
Issue 157: ReadOnlyState support

mcoblenz:
Added support for "this" arguments that have no permission specified, in which case ReadOnlyState is inferred. "this" arguments that are missing are still treated as Shared. 

This change includes a few small improvements that I made along the way, including removal of the availableIn field from transactions and a fix to the parser to no longer use "THIS" as a contract name.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
@@ -949,6 +949,10 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
 
                 val oldType = context.thisType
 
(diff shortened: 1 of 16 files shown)


src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -854,7 +864,13 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
                 receiverType match {
                     case typ: NonPrimitiveType => {
                         val newType = invokable.thisFinalType(typ.contractName)
```
pkoronkevich: thisFinalType(contractName: String) def can now be removed

PR #180, opened 2018-08-01
Issue 176 and 179

pkoronkevich:
Fixes the issue of invoking a transaction (or constructor) on an Owned reference and taking ownership, even if the transaction specifies that it only needs an Unowned reference. 

Also addresses `throw` as a return statement issue.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/typecheck/Error.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/typecheck/Error.scala
@@ -257,6 +257,6 @@ case class InconsistentTypeAssignmentError(declaredType: ObsidianType, actualTyp
     val msg: String = s"Cannot assign a value of type $actualType to a variable of type $declaredType."
 }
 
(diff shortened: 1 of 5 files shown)


src/main/scala/edu/cmu/cs/obsidian/typecheck/Error.scala
```diff
@@ -255,4 +255,8 @@ case class InconsistentContractTypeError(declaredContractName: String, actualCon
 
 case class InconsistentTypeAssignmentError(declaredType: ObsidianType, actualType: ObsidianType) extends Error {
     val msg: String = s"Cannot assign a value of type $actualType to a variable of type $declaredType."
+}
+
+case class ArgumentSubtypingError(tName: String, t1: ObsidianType, t2: ObsidianType) extends Error {
+    val msg: String = s"Found type '$t1' as an argument to '$tName', but expected something of type '$t2'."
```
mcoblenz: Maybe also say which argument it is.

PR #201, opened 2019-02-19
Fabric integration

YBthebest:
Generates the correct structure we need for Fabric integration, including the right build.gradle file and the .java and outerClass.java files. Includes comments to explain work.
Also removed the .jar generation if we are generating real chaincode.


src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -210,6 +212,54 @@ object Main {
         f.delete()
     }
 
+    def generateFabricCode(mainName: String, outputPath: Option[String], srcDir: Path): Unit = {
+        try {
+            //what we need to do now is move the .java class and the outterclass to a different folder
```
mcoblenz: Fix typo ("outterclass")

src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -210,6 +212,54 @@ object Main {
...
+            /* if an output path is specified, use it; otherwise, use working directory */
+            val outputPath_temp = outputPath match {
+                case Some(p) =>
+                    val path = Paths.get(p + mainName)
+                    /* create the dir if it doesn't exist */
+                    Files.createDirectories(path)
+                    path
+                case None =>
+                    val path = Paths.get(mainName)
```
mcoblenz: De-duplicate this; only the assignment to `path` differs between branches.

src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -210,6 +212,54 @@ object Main {
...
+                    path
+            }
+
+            //copy the content of the fabric/java/ folder into a folder with the class name
+            val copyFabricFolderInvocation: String =
+                "cp -R fabric/java/ " + outputPath_temp
+            copyFabricFolderInvocation.!
+
+            //copy both the java and outerClass.java files to the Fabric folder
+            val javaSourceLocation = Paths.get(srcDir + "/edu/cmu/cs/obsidian/generated_code/" + mainName + ".java")
```
mcoblenz: Rather than concatenating, use the multi-argument version of Paths.get:

```
public static Path get(String first,
                       String... more)
```

src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -210,6 +212,54 @@ object Main {
...
+            val outerJavaTargetLocation = Paths.get(mainName + "/src/main/java/org/hyperledger/fabric/example/" + protobufOuterClassNameForClass(mainName) + ".java")
+            Files.copy(outerJavaSourceLocation, outerJavaTargetLocation, StandardCopyOption.REPLACE_EXISTING)
+
+            //place the correct class name in the build.gradle
+            val replaceClassNameInGradleBuild: String =
+                "sed -i .backup s/{{CLASS_NAME}}/" + mainName + "/g " + mainName + "/build.gradle"
+            //sed automatically creates a backup of the original file, has to be deleted
+            val deleteSedBackupFile: String =
+                "rm " + mainName + "/build.gradle.backup"
+            //.keep file is only there since we need it to get the original Fabric structure on Git
```
mcoblenz: Does it hurt anything to leave the .keep file in place?

src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -346,9 +397,9 @@ object Main {
                 } catch {
                     case e: Throwable => println("Error running protoc: " + e)
                 }
+                generateFabricCode(mainName, options.outputPath, srcDir)
```
mcoblenz: Only in non-mock mode?

src/main/scala/edu/cmu/cs/obsidian/Main.scala
YBthebest: I don't think it technically hurts anything. I figured it was cleaner to get rid of it once and for all, but I can leave it there.

src/main/scala/edu/cmu/cs/obsidian/Main.scala
YBthebest: Yep, that was during testing, good catch. Thanks.

PR #205, opened 2019-03-21
Get guid

shell769324:
Make a script getGuid that allows user to retrieve the guid of the object returned from the latest invocation.


network-framework/scripts/invoke.sh
```diff
@@ -74,18 +74,33 @@ parseParameters() {
...
-set +x
-cat log.txt
-verifyResult $res "Invoke execution on $PEERS failed "
-echo "===================== Invoke transaction successful on $PEERS on channel '$CHANNEL_NAME' ===================== "
-echo
\ No newline at end of file
+if [ $isQuiet -eq 1 ]; then
+  peer chaincode invoke -o orderer.example.com:7050 --tls $CORE_PEER_TLS_ENABLED --cafile $ORDERER_CA -C $CHANNEL_NAME -n mycc $HARD_PEER_CONN_PARMS -c $PARAMS >&log.txt
+  res=$?
+  verifyResult $res "Invoke execution on $PEERS failed "
```
mcoblenz: In quiet mode, please translate failures to an exit status of 1 (instead of 0).

shell769324: In case of failure, verifyResult will exit with status 1.

mcoblenz: I see, okay. Carry on, then.

PR #208, opened 2019-04-05
Enable integration tests framework

shell769324:
Refurbish integration test framework
Allow travis to run tests on hyperleder fabric
Write a small int container test script as a demo
Delete unused manifest file


resources/tests/endToEndTests/IntContainerTest.sh
```diff
@@ -0,0 +1,27 @@
...
+
+
+testVals=(5 7 8 9 10)
+for ((i = 0; i < 4; i++)); do
+    echo -n "./invoke.sh -q setX ${testVals[i + 1]}"
+    res=`./invoke.sh -q setX ${testVals[i + 1]}`
+    sleep 2
+    if [[ ${res} -ne testVals[i] ]] ; then
+        echo
+        echo "The old value of IntContainer is ${testVals[i]}, but it returns $res"
```
mcoblenz: Fix grammar in error message.

network-framework/channel-artifacts/holder.txt
```diff
@@ -0,0 +1 @@
+upload this artifacts
```
mcoblenz: What is this? Do we need it? If so, add an explanation (and fix grammar).

travis_specific/install-bin.sh
```diff
@@ -0,0 +1,143 @@
+#!/bin/bash
```
mcoblenz: Do we really want to have this file here instead of downloading it every time? If so, why is it necessary? I'd rather fetch it so we remain compatible with future Fabric releases.

network-framework/scripts/invoke.sh
```diff
@@ -16,6 +16,15 @@ PROTO_PATH="${CC_SRC_PATH}protos/"
 CC_NAME=$(basename $CC_SRC_PATH)
 
 
+verifyResult() {
```
mcoblenz: Remove duplicate copy of this function.

network-framework/channel-artifacts/.gitkeep
```diff
@@ -0,0 +1 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
mcoblenz: Still need to address this.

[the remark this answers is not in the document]
shell769324: channel-artifacts is an empty folder that will store some files when fabric runs. I don't have a better way to push an empty directory, and putting a gitkeep seems to be the convention. I will delete this file since the folder already exists on gitHub.

travis_specific/install-bin.sh
shell769324: This file has been deleted since then. The latest version doesn't have it.

PR #212, opened 2019-04-11
Fixes #206 by adding context for false branch in IfInState TypeChecker

### LINKED ISSUE #206: Make dynamic state check type reasoning smarter
```
 transaction t4(LightSwitch@Owned s) {
        if (s in On) {
             s.turnOff();
        }

        [s @ Off];
}
```
The assertion above currently fails because the type checker isn't smart enough. Let's make it succeed (assuming that the only possible states are On and Off).

YBthebest:
We add a different context that expands the Owner states into all possible states and removes the one state we checked for in the true branch context. This fixes #206

### DIFF
--- a/resources/tests/type_checker_tests/InState.obs
+++ b/resources/tests/type_checker_tests/InState.obs
@@ -51,13 +51,13 @@ main contract InState {
              s.turnOff();
         }
 
-        // TODO: Make the type checker smart enough for this assertion to pass.
-        // [s @ Off];
+        [s @ Off];
 
         if (s in Off) {
              s.turnOn();
         }
 
+        // ERROR: s is On now, we know this specifically
         [s @ Owned];
     }
 }
--- a/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
@@ -1488,8 +1488,14 @@ private def checkStatement(
                                     np.permission match {
                                         case Owned() =>
                                             val newType = StateType(np.contractName, Set(state._1), np.isRemote)
-                                            val typeFalse = StateType(np.contractName, allStates - state._1, np.isRemote)
-                                            (contextPrime.updated(x, newType).updatedMakingVariableVal(x), contextPrime.updated(x, typeFalse).updatedMakingVariableVal(x))
+                                            t match {
+                                                case StateType(_, specificStates, _) =>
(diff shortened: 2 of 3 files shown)


src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -1468,17 +1488,19 @@ private def checkStatement(
                                     np.permission match {
                                         case Owned() =>
                                             val newType = StateType(np.contractName, Set(state._1), np.isRemote)
-                                            contextPrime.updated(x, newType).updatedMakingVariableVal(x)
-                                        case Unowned() => contextPrime
+                                            val typeFalse = StateType(np.contractName, allStates - state._1, np.isRemote)
```
mcoblenz: This is close, but what if this is a StateType (not just any old NonPrimitiveType)? Then we know something more specific about the possible states than just that it's one of allStates.

resources/tests/type_checker_tests/InState.obs
```diff
@@ -51,13 +51,13 @@ main contract InState {
...
 
-        // TODO: Make the type checker smart enough for this assertion to pass.
-        // [s @ Off];
+        [s @ Off];
 
         if (s in Off) {
              s.turnOn();
         }
 
+        // ERROR: s is On now, we know this specifically
```
mcoblenz: Need to update the unit tests to expect this error.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
YBthebest: @mcoblenz I put allStates - state._1, isn't that what you are referring too?

resources/tests/type_checker_tests/InState.obs
YBthebest: Where can I find the unit tests?

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
mcoblenz: But allStates is all the states in the contract. If the current type is a particular set of states, you should use that instead.

resources/tests/type_checker_tests/InState.obs
mcoblenz: Look at TypeCheckerTests.scala. Or run sbt test and you'll see what fails.

PR #214, opened 2019-04-23
Add support to display error messages

### LINKED ISSUE #211: Report server error messages to clients
When a server invocation fails, doTransaction on the client should throw an exception that includes the error message.

YBthebest:
Fixes #211 by throwing a `ChaincodeClientTransactionFailedException` error with the value returned by invoke.sh if the invoke.sh script has an exit value different to 0.


YBthebest:
Issues with building the runtime locally due to Java versions divergence. Will fix this PR.

PR #215, opened 2019-04-23
Remove .jar dependencies and add required one to the build.sbt to imp…

### LINKED ISSUE #187: Properly find libs regardless of working directory
We should support a working directory that is independent of the compiler's location. The working directory should be wherever the code is.

### LINKED ISSUE #213: Remove `invokeJavac` method from Main
We don't invoke javac directly anymore.

YBthebest:
Fixes #187 and fixes #213  by removing the external .jar files from the project and adding the required dependencies to the build.sbt and by removing the invokeJavac method from Main


YBthebest:
Also fixes #213

PR #220, opened 2019-05-17
Added check for duplicate contract

### LINKED ISSUE #116: Handle duplicate contracts
**Describe the bug**
If a file has duplicate contracts, an "assertion failed" exception is raised from the type checker, which doesn't provide a clear indication of why it was raised.

**To Reproduce**
Compile a single file that has duplicate contracts.

**Expected behavior**
A proper error message warning the user of duplicate contracts.

YBthebest:
Fixes #116. Checks if multiple contracts have the same name, and if so, stops the compiling with an error message indicating the name of the culprit contract.


mcoblenz:
Do we need to stop compiling, or can we ignore the duplicate and continue?

YBthebest:
We can ignore it. I just pushed to display a warning instead of an error.

PR #223, opened 2019-05-31
Archiving Fixes in Code Generator

ReedOei:
This pull request fixes two issues related to the archiving of objects:

1. Doing a dynamic state check does not load the state from the archive, meaning that if it is the first thing a transaction does will an object, the code will crash with a NullPointerException. To fix this, there are now two versions of the `getState` method generate: the first of these takes `SerializedState __st` as a parameter, while the second is identical the old `getState`. The first version will always call `__restoreObject` on the object, ensuring that the object is loaded before checking its state. The second version exists for the cases in which we can be sure that the object will be loaded before it is called, so passing around the `__st` parameter would be a waste and only complicate the generated code. 

2. Previously, objects were not marked as modified just because the state changed. Then if the only thing a transaction does is transition the state of an object, without modifying any of its fields, it will not be marked as modified and the changes will not be saved. I fixed this by generating the line `__modified = true;` whenever there is a state transition.


mcoblenz:
You'll need to address the failing tests.

ReedOei:
@mcoblenz The reason the tests are failing is the change to `__getState` requires that it throws an `InvalidProtocolBufferException`, but the rest of the methods that call it do not. I can add it to all of the necessary methods, but I will also need to change the runtime library.

I also considered just adding the `__restoreObject` call before the uses of `__getState` in the generated that may require loading, but was worried about potentially missing some cases.

What do you think is the best way to do this?

mcoblenz:
Probably we should propagate the exception up (and declare throwing it in that method), since the alternative is to eat it.

ReedOei:
@mcoblenz I believe I have fixed the issue, but since the runtime library is pulled from the jar on `master`, Travis is still using the old version, causing the build to fail. When I build it locally using the runtime jar that is in `compiler-archive-fixes`, I am able to build things without errors. Not sure how to get the Travis build to pass, outside of pushing the new jar to master.

mcoblenz:
This looks good. You won't be able to get the CI tests passing until you merge, so go ahead and merge/squash and then make sure CI is OK afterward.

PR #229, opened 2019-06-05
Require constructors to include a permission

### LINKED ISSUE #225: Compiler doesn't know state after constructor
In the following contract, the compiler doesn't seem to know what state an `X` is in after being created, but it seems like it should:

```
main contract X {
    state S1;
    state S2;

    X() {
        ->S1;
    }

    transaction f(X@S1 >> S2 this) {}

    transaction g() {
        X x = new X();
        x.f(); // error occurs here
    }
}
```

Compiling this gives:

```
test.obs 15.11: Cannot invoke f on a receiver of type X@Owned; a receiver of type X@S1 is required.
```

EDIT: It seems you can change the constructor to read `X@S1() { ... }` and it will work. Still, maybe there should be some warning for this. Or, perhaps, since the constructor must put the object into some state, the @S1 part should be required.

ReedOei:
This pull request changes constructor syntax so that:

```
main contract C {
    state S1;
    C() { ->S1; }
}
```

is no longer allowed. It fails with the error:

```
Error: Missing a permission on constructor for C (e.g., C@Owned, C@State, where State is a valid state for C). at 4:1 in test.obs
```

Instead, it must be:

```
main contract C {
    state S1;
    C@S1 { ->S1; }
}
```

This closes #225.


ReedOei:
@mcoblenz I have made the requested changes.

PR #230, opened 2019-06-05
Allow overloading constructors

### LINKED ISSUE #172: Generating code with the same constructors
**Describe the bug**
When a contract is defined with multiple constructors that take the same arguments that aren't subtypes (i.e. a C@S1 vs C@S2), it passes the type checker but the constructors generated in Java take the same arguments.

**To Reproduce**
```
main contract A {

    transaction makeB() {
        B@Shared b = new B(new C());
    }
}

contract B {

    B@Shared(C@S1 s1c) {}
    B@Shared(C@S2 s2c) {}
}

contract C {
    state S1;
    state S2;

    C@S1() {
        ->S1;
    }
}
```
results in
`B.java:14: error: constructor B( C ) already defined in class B public B(C s2c) `

ReedOei:
Previously, only one constructor was allowed. This pull request allows for multiple constructors, as long as their parameters have different types (including state).

For example, the below code used to fail to compile. However, now when we compile:

```
contract B { 
    int x;

    B@Owned(C@(S1 | S3) s1c) {
        x = 1;
    }   

    B@Owned(C@S2 s2c) {
        x = 2;
    }
}
```

the compiler generates:

```java
private void new_B(C s1c, SerializationState __st)
    throws com.google.protobuf.InvalidProtocolBufferException, ObsidianRevertException
{
    C s2c = s1c;
    if (s1c.getState(__st).toString().equals("S1")||s1c.getState(__st).toString().equals("S3")) {
        __constructorReturnsOwned = true;
        this.x = java.math.BigInteger.valueOf(1);
        __modified = true;
    } else {
        if (s2c.getState(__st).toString().equals("S2")) {
            __constructorReturnsOwned = true;
            this.x = java.math.BigInteger.valueOf(2);
            __modified = true;
        }
    }
    __guid = __st.getUUIDFactory().newUUID().toString();
    __modified = true;
    __loaded = true;
    __st.putEntry(__guid, this);
}
```

This closes #172.


ReedOei:
@mcoblenz If we're going to allow this for constructors, perhaps we should also allow it for transactions in general?

ReedOei:
@mcoblenz I have made the requested changes.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -928,6 +929,130 @@ class CodeGen (val target: Target, table: SymbolTable) {
...
+
+        // Build the conditions to check whether transaction's body to run
+        val conditions = decls.map(constructor => (constructor, allExpr(constructor.args.flatMap(buildConditions))))
+
+        def setOwned(block: JBlock, constructor: Constructor): JBlock =
+            block.assign(JExpr.ref(constructorReturnsOwnedFieldName), JExpr.lit(constructor.resultType.isOwned))
+
+        def addDispatchCondition(ifSt: JConditional)(info: (Constructor, IJExpression)): Unit = info match {
+            case (decl, cond) =>
+                translateBody(setOwned(ifSt._elseif(cond)._then(), decl), decl.body, translationContext, localContext)
```
mcoblenz: Separate these lines; don't pass side-effecting method call results as arguments.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -2127,17 +2127,71 @@ private def checkStatement(
...
 
-        // TODO: enable multiple constructors. by adding appropriate dispatch logic in init. For now, we only support one constructor per contract.
-        if (constructors.length > 1) {
+        if (contract.isMain && constructors.length > 1) {
             logError(contract, MultipleConstructorsError(contract.name))
         }
 
-        val constructorsByArgTypes = constructors.groupBy(c => c.args.map(_.typIn.topPermissionType))
-        val matchingConstructors = constructorsByArgTypes.filter(_._2.size > 1)
+        def baseTypeName(arg: VariableDeclWithSpec): String =
```
mcoblenz: This looks like it belongs as a method of VarDeclWithSpec.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -2127,17 +2127,71 @@ private def checkStatement(
...
+        def states(t: NonPrimitiveType): Set[String]=
+            t match {
+                case ContractReferenceType(contractType, permission, isRemote) =>
+                    table.lookupContract(contractType.contractName).map(_.possibleStates).getOrElse(Set())
+                case InterfaceContractType(name, simpleType) => states(simpleType)
+                case StateType(contractName, stateNames, isRemote) => stateNames
+            }
+
+        // Two types are distinguishable iff they are both owned and have non-overlapping state lists
+        def distinguishable(args: (ObsidianType, ObsidianType)): Option[String] = {
```
mcoblenz: The name and comment suggest that this should return a Boolean, but it does not. Please update the name and comment accordingly.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -2127,17 +2127,71 @@ private def checkStatement(
...
+                // This case should never happen, since we already grouped the constructor arguments on type
+                case _ => assert(false, s"Unexpected distinguishability test: $args"); None
+            }
+        }
+
+        // Returns the first argument pair that is ambiguous, if EVERY argument pair is ambiguous
+        // If any pair is distinguisable, that means the both constructors are themselves distinguishable
+        def distinguish(constructor: Constructor, other: Constructor): Option[(String, (VariableDeclWithSpec, VariableDeclWithSpec))] =
+            constructor.args.zip(other.args).map(args =>
+                    distinguishable((args._1.typIn, args._2.typIn)).flatMap(ex => Some(ex, args))
```
mcoblenz: This is super hard to read. To make this easier, break it into pieces and give the pieces names (via local variables).

PR #241, opened 2019-06-14
Remove erroneous return statement check on interface transactions

### LINKED ISSUE #240: Interfaces improperly require return statements
The following fails to compile
```
interface IHash {
    transaction sha256(string message) returns string;
}

contract Hash implements IHash with java : edu.cmu.cs.obsidian.stdlib.Hash;

main contract Test {}
```

with the error:
```
 2.5: 'sha256' specifies a return type, but no return value is given.
```

Presumably this should give no error, since interfaces don't implement methods.

ReedOei:
In the case of interfaces, we should not require that methods be checked for a return statement. This fix simply ignores this check if the contract containing the transaction is an interface.

This closes #240.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
@@ -1853,22 +1853,27 @@ private def checkStatement(
 
         checkForUnusedStateInitializers(outputContext)
 
-        if (tx.retType.isDefined & !hasReturnStatement(tx, tx.body)) {
-            val ast =
-                if (tx.body.length > 0) {
-                    tx.body.last
+        lexicallyInsideOf.contract match {
+            case ObsidianContractImpl(modifiers, name, declarations, transitions, isInterface, sp) =>
+                // Don't need to check interface methods to make sure they return
+                // Use & instead of && to ensure that all conditions evaluate (hasReturn checks for multiple things)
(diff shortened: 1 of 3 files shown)


src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -1853,22 +1853,27 @@ private def checkStatement(
...
-            val ast =
-                if (tx.body.length > 0) {
-                    tx.body.last
-                }
-                else {
-                    tx
+        lexicallyInsideOf.contract match {
+            case ObsidianContractImpl(modifiers, name, declarations, transitions, isInterface, sp) =>
+                // Don't need to check interface methods to make sure they return
+                // Use & instead of && to ensure that all conditions evaluate (hasReturn checks for multiple things)
```
mcoblenz: Interesting, but it might be better to just change the order so hasReturnStatement goes first. Then you can use &&.

ReedOei: That's how it was originally, so I just left it there (after changing it to `&&`, having thought it was a typo, and subsequently spending some time debugging it).

mcoblenz: I see. Still, it might be worth changing, just for uniformity and simplicity.

PR #255, opened 2019-06-27
Improve build script to work from any directory

ReedOei:
Changes as discussed:

- Build script can work from any directory
- Will automatically build jar if it does not exist
- Can pass compiler flags in addition to file path now


ReedOei:
@mcoblenz Builds seem to be failing on Travis semi-frequently with errors due to downloading dependencies. Not sure what to do, other than hope it will resolve itself...

PR #260, opened 2019-06-28
Import search path improvements

ReedOei:
As discussed, the `import` statement has been improved: there are now three search paths, which are searched in the order below (cf. before only 3. was supported):

1. Relative to the file being compiled
2. The standard library path: this is read from the environment variable `OBSIDIAN_STDLIB`, or defaults to "$compiler_path/Obsidian_Runtime/src/main/java/Runtime/edu/cmu/cs/obsidian/stdlib"
3. The compiler path: read from "TRAVIS_BUILD_DIR", "OBSIDIAN_COMPILER_DIR", or defaulting to the current working directory.


src/main/scala/edu/cmu/cs/obsidian/Main.scala
```diff
@@ -189,16 +192,7 @@ object Main {
...
-            if (compilerDir == null) {
-                // TODO: package up the compiler as a jar file and use a path relative to that
-                // Requiring the working dir to be the compiler's directory is inconvenient.
-                compilerDir = System.getenv("OBSIDIAN_COMPILER_DIR")
-                if (compilerDir == null) {
-                    compilerDir = Paths.get("").toAbsolutePath().toString
-                }
-            }
-            val fabricPath = Paths.get(compilerDir, "fabric", "java")
+            val fabricPath = ImportProcessor.compilerPath().resolve("fabric").resolve("java")
```
mcoblenz: This seems like a weird place to put compilerPath. Maybe the method should be defined right here (in Main) and then passed as an argument when instantiating the ImportProcessor?

src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala
```diff
@@ -12,11 +13,37 @@ object ImportProcessor {
 
     class ImportException (val message : String) extends Exception {}
 
+    def compilerPath(): Path =
+        List("TRAVIS_BUILD_DIR","OBSIDIAN_COMPILER_DIR")
```
mcoblenz: We should document all these environment variable names somewhere, but I realize there's aren't new in your change.

src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala
```diff
@@ -12,11 +13,37 @@ object ImportProcessor {
...
+            .map(System.getenv)
+            .find(_ != null)
+            .map(Paths.get(_).toAbsolutePath)
+            .getOrElse(Paths.get("").toAbsolutePath)
+
+    def standardLibraryPath(): Path = {
+        val envPathStr = System.getenv("OBSIDIAN_STDLIB")
```
mcoblenz: In what cases is this necessary? I guess I'd expect the stdlib to get packaged up into the Jar file with the compiler, and then the path will always be relative to the compiler's jar file.

src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala
ReedOei: Is there some existing place I should do this, or do we just want to create an issue for this or something?

src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala
ReedOei: If the stdlib is packaged in the jar file it won't be accessible like this anyway (we'd have to decompress the jar to access it). If it's just included alongside the jar file (at the same relative path), the default path solution still works, and if not, then the environment variable allows for flexibility in putting the standard library wherever we want.

ReedOei:
@mcoblenz I have made the changes we discussed. I also removed the `TRAVIS_BUILD_DIR` environment variable, since it seems to never actually be set.

src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala
```diff
@@ -13,32 +14,46 @@ object ImportProcessor {
...
-    private def searchPaths(importingFile: String): Seq[Path] =
-        List(relativePath(importingFile), standardLibraryPath(), compilerPath())
+    private def compilerPath(fileName: String): Option[(String, InputStream)] =
+        resolvePathAndInputStream(Main.compilerPath(), fileName)
 
-    def searchImportFile(importingFile: String, fileName: String): Option[Path] =
-        searchPaths(importingFile)
-            .map(_.resolve(fileName))
-            .find(Files.exists(_))
+    // TODO: Write about the environment variables in the README
```
mcoblenz: This TODO is done, right?

PR #262, opened 2019-06-28
Add exceptions to generated stub transactions

### LINKED ISSUE #259: Generated client code missing exceptions
Some of the stub code the Obsidian compiler generates does not compile. In particular, for the Insurance case study, the code compiles (with Obsidian), but has the following error when built with Gradle: 

```
> Task :compileJava FAILED
/home/roei/ObsidianImports/obs_output/Insurer/src/main/java/org/hyperledger/fabric/example/Money__Stub__.java:26: error: unreported exception InvalidProtocolBufferException; must be caught or declared to be thrown
        argArray.add(Base64 .getEncoder().encodeToString(m.__wrappedArchiveBytes()));
                                                                                ^
1 error

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler error output for details.

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.
```

ReedOei:
This closes #259.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
@@ -380,8 +380,8 @@ class CodeGen (val target: Target, table: SymbolTable) {
         }
 
         val meth: JMethod = newClass.method(JMod.PUBLIC, javaRetType, transaction.name)
+        addTransactionExceptions(meth, translationContext)
         meth._throws(model.directClass("edu.cmu.cs.obsidian.client.ChaincodeClientAbortTransactionException"))
-
 
         var argExpressions: List[IJExpression] = Nil
         /* add args */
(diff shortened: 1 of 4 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -380,9 +380,9 @@ class CodeGen (val target: Target, table: SymbolTable) {
         }
 
         val meth: JMethod = newClass.method(JMod.PUBLIC, javaRetType, transaction.name)
+        addTransactionExceptions(meth, translationContext)
         meth._throws(model.directClass("edu.cmu.cs.obsidian.client.ChaincodeClientAbortTransactionException"))
```
mcoblenz: Can we remove this _throws line now?

ReedOei: Not really, though of course we could move it somewhere, but I'm not sure what would be better; `addTransactionExceptions` only adds the `ChaincodeClientAbortTransactionException` to the throws list in the case that the target is a Client rather than a Server, but that's not the case here.

PR #266, opened 2019-07-16
Give an error when passing owned references as arguments to transacti…

mcoblenz:
…ons that are going to discard ownership.

In a future change, I'll add the remaining cases to AllPermissions.obs (passing shared and unowned arguments).

### DIFF
--- a/resources/tests/type_checker_tests/AllPermissions.obs
+++ b/resources/tests/type_checker_tests/AllPermissions.obs
@@ -8,9 +8,13 @@ main asset contract AllPermissions {
 
     transaction t3(AllPermissions @ Unowned this) {}
 
-    transaction t4(AllPermissions @ Owned >> Shared this) {}
+    transaction t4(AllPermissions @ Owned >> Shared this) {
+        disown this;
+    }
 
-    transaction t5(AllPermissions @ Owned >> Unowned this) {}
+    transaction t5(AllPermissions @ Owned >> Unowned this) {
+        disown this;
+    }
 
     transaction t6(AllPermissions @ Shared >> Owned this) {
         // Error; no way to make this typecheck.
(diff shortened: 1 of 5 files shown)


resources/tests/type_checker_tests/AllPermissions.obs
```diff
@@ -0,0 +1,167 @@
...
+        x3.t3();
+        [x3 @ Owned]; // Special exception!
+        disown x3;
+
+        AllPermissions x4 = new AllPermissions();
+        x4.t4();
+        [x4 @ Shared];
+
+        AllPermissions x5 = new AllPermissions();
+        x5.t5();
```
ReedOei: Shouldn't there be some error (either here or in t5), since we lose ownership of x5 (an asset) without a disown?

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
```diff
@@ -100,7 +100,11 @@ case class VariableDecl(typ: ObsidianType, varName: String) extends Statement
 case class VariableDeclWithInit(typ: ObsidianType, varName: String, e: Expression) extends Statement
 
 case class VariableDeclWithSpec(typIn: ObsidianType, typOut: ObsidianType, varName: String) extends Statement
-
+ {
+     override def toString: String = {
```
ReedOei: This could just be `override def toString: String = varName`

resources/tests/type_checker_tests/AllPermissions.obs
mcoblenz: Good catch. I'll fix this (the error should be in t5).

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
mcoblenz: OK, sure.

PR #270, opened 2019-07-29
Obsidian Generics

ReedOei:
Adds generics to Obsidian. See `resources/tests/type_checker_tests` and `resources/tests/compilerTests` for examples.


src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -268,65 +265,108 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
...
+                    case (StateType(c1, ss1, _), ContractReferenceType(c2, c2p, _)) =>
+                        contractIsSubtype(table, c1, c2, isThis)
                     case _ => false
                 }
-                if (!mainSubtype) Some(SubtypingError(t1, t2, isThis))
-                // TODO: make sure this is unnecessary: else if (!modifierSubtype) Some(OwnershipSubtypingError(t1, t2))
-                else None
-            case _ => Some(SubtypingError(t1, t2, isThis))
+
+            case (p: PrimitiveType, np: NonPrimitiveType) => np.contractName == "Top"
```
mcoblenz: Avoid hardcoding "Top" in various places; define an identifier for it somewhere (probably in ObsidianTypes).

mcoblenz: Actually I suggest calling this "Contract", analogous to Java's "Object".

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -1019,37 +1043,148 @@ class CodeGen (val target: Target, table: SymbolTable) {
...
+        target.typeParams().last.bound(boundClass(typeVar, param.bound))
+    }
+
+    def obsidianSerialized: AbstractJClass =
+        model.directClass("edu.cmu.cs.obsidian.chaincode.ObsidianSerialized")
+
+    def setType(): AbstractJClass = model.directClass("java.util.Set")
+    def hashSetType(): AbstractJClass = model.directClass("java.util.HashSet")
+
+    def javaString(): AbstractJClass = model.directClass("java.lang.String")
```
mcoblenz: Call this javaStringType to match the above two.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGen.scala
```diff
@@ -120,28 +121,37 @@ class CodeGen (val target: Target, table: SymbolTable) {
             }
         }
 
-        /* match on c to check which type of contract it is */
-        for (c <- program.contracts) {
-            // TODO : generate code for interfaces (issue #117)
```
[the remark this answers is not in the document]
ReedOei: Yes, we now generate Java interfaces for Obsidian interfaces.

resources/tests/type_checker_tests/StaticAsserts.obs
```diff
@@ -23,6 +23,8 @@ main contract C {
 
         [ow @ Unowned]; // should fail
 
-        [ow @ Owned | Unowned];
+        // TODO: Not sure why we want this.
+        // In the interest of making thing more uniform, I ended up accidentally disallowing this---I can put it back if necessary.
+        // [ow @ Owned | Unowned];
```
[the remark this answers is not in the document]
ReedOei: Usually we either have a permission, a set of states, or a permission variable. This breaks the pattern by allowing a list of permissions. Additionally, since you'll always know which permission the reference has at compile time, it doesn't seem useful to write anything other than the actual permission it has.

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
```diff
@@ -48,79 +58,266 @@ sealed abstract class InvokableDeclaration() extends Declaration {
...
     override val toString = name
+
+    override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): ReferenceIdentifier = this
 }
 
-case class NumLiteral(value: Int) extends Expression
-case class StringLiteral(value: String) extends Expression
-case class TrueLiteral() extends Expression
-case class FalseLiteral() extends Expression
+case class NumLiteral(value: Int) extends Expression {
```
mcoblenz: This is a lot of duplicated code. The following might be a little better…

````
abstract case class BinaryExpression(e1: Expression, e2: Expression) extends Expression {
    def make(newE1: Expression, newE2: Expression): BinaryExpression

    override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): BinaryExpression = {
        make(e1.substitute(genericParams, actualParams), e2.substitute(genericParams, actualParams)).setLoc(this)
    }
}

case class Conjunction(override e1: Expression, override e2: Expression) extends BinaryExpression(e1, e2) {
    def make(newE1: Expression, newE2: Expression): BinaryExpression = Conjunction(newE1, newE2)
}
````
etc.

Comment (mcoblenz) [reply src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala]: I wonder if there's a Scala mechanism that would let us cut out all those `make` methods, but if so, it's not apparent to me what it is.

File: src/main/scala/edu/cmu/cs/obsidian/parser/ImportProcessor.scala [hunk 6]
```diff
@@ -109,12 +109,11 @@ object ImportProcessor {
 
     /*check this */
     def filterTags(contracts: Seq[Contract]): Seq[Contract] = {
-        contracts.map(c => c match {
```
mcoblenz: I'll accept this change, but minor changes like this make PRs harder to review and make commit logs harder to understand. It's inconvenient, but it's better form to propose these kinds of changes separately.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -91,42 +91,47 @@ object Parser extends Parsers {
...
         val intPrim = IntT() ^^ { t => IntType().setLoc(t) }
         val boolPrim = BoolT() ^^ { t => BoolType().setLoc(t) }
         val stringPrim = StringT() ^^ { t => StringType().setLoc(t) }
 
         parseNonPrimitive | intPrim | boolPrim | stringPrim
     }
 
-    private def extractTypeFromPermission(permission: Option[~[Token, Seq[Identifier]]], name: String, isRemote: Boolean, defaultOwned: Boolean): NonPrimitiveType = {
+    private def extractTypeFromPermission(permission: Option[~[Token, Seq[Identifier]]],
+                                          name: String, genericParams: Seq[ObsidianType],
```
mcoblenz: One parameter per line if you're going to break it up, please.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -91,42 +91,47 @@ object Parser extends Parsers {
...
+
+        val contractType = ContractType(name, genericParams)
+
         permission match {
-            case None => ContractReferenceType(ContractType(name), defaultPermission, isRemote)
+            case None => ContractReferenceType(contractType, defaultPermission, isRemote)
             case Some(_ ~ permissionIdentSeq) =>
                 if (permissionIdentSeq.size == 1) {
                     val thePermissionOrState = permissionIdentSeq.head
-                    val permission = resolvePermission(thePermissionOrState._1)
```
mcoblenz: This refactor wasn't really necessary. In general, unnecessary changes make PRs harder to review. This change in particular takes work that was broken into two steps (for readability) and re-combines it into one, which is harder to read.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -601,12 +641,86 @@ object Parser extends Parsers {
...
+                val params = optParams.map(_._1._2).getOrElse(List())
+                ContractType(name._1, params)
+        }
+    }
+
+    private def parseWhereImplements: Parser[Option[ContractType]] = {
+        WhereT() ~ parseId ~ ImplementsT() ~ parseType ^^ {
+            case _ ~ typeVar ~ _ ~ typeBound => typeBound match {
+                case np: NonPrimitiveType => Some(np.contractType)
+                case _ => failure("Contract cannot implement primitive type."); None
```
mcoblenz: If you give a parse error, the whole compilation process will stop. Can you give a type error in the typechecker instead?

src/main/scala/edu/cmu/cs/obsidian/parser/SymbolTable.scala
```diff
@@ -90,6 +89,21 @@ class ContractTable(
         val contract: Contract,
         symbolTable: SymbolTable,
         parentContract: Option[ContractTable]) extends DeclarationTable {
+
+    /**
+      * Constructs a new contract table with the current contract updated to
```
mcoblenz: updated to what?

src/main/scala/edu/cmu/cs/obsidian/parser/SymbolTable.scala
```diff
@@ -129,24 +165,24 @@ class ContractTable(
...
-                    case (_, Some(ct)) => Some(ct)
-                    case (Some(ct), _) => Some(ct)
-                    case _ => None
-                }
-        if (localLookupResult.isDefined) {
-            localLookupResult
-        } // otherwise nothing was found
-        else None
-
+        if (name == this.name) {
```
mcoblenz: Yeah, this is better, thanks.

src/main/scala/edu/cmu/cs/obsidian/typecheck/Error.scala
```diff
@@ -360,3 +361,94 @@ case class ConstructorAnnotationMissingError(contractName: String) extends Error
...
+                                    isSubperm: Boolean) extends Error {
+    private val isSubpermStr: String =
+        if (isSubperm) {
+            "is a subpermission. The then-branch will always be taken"
+        } else {
+            "is not a subpermission. The else-branch will always be taken"
+        }
+
+    override val msg: String =
+        s"Redundant permission check. Testing for permission $testPermission, but actual permission is $isSubperm, " +
```
mcoblenz: Redundant "is".

src/main/scala/edu/cmu/cs/obsidian/typecheck/Error.scala
```diff
@@ -360,3 +361,94 @@ case class ConstructorAnnotationMissingError(contractName: String) extends Error
...
+}
+
+case class AssetStateImplError(implState: State, interfaceState: State) extends Error {
+    override val msg: String =
+        s"State ${implState.name} is an asset state but it implements a state which is not an asset state."
+}
+
+case class BadFFIInterfaceBoundError(name: String) extends Error {
+    override val msg: String =
+        s"Cannot override '$name' because it is an FFI contract."
```
mcoblenz: Not sure if users will know what "FFI" means. Maybe "because it is implemented in Java, not Obsidian."

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -268,65 +265,125 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
...
 
     /* true iff [t1 <: t2] */
-    private def isSubtype(t1: ObsidianType, t2: ObsidianType, isThis: Boolean): Option[Error] = {
-        (t1, t2) match {
-            case (BottomType(), _) => None
-            case (_, BottomType()) => None
-            case (IntType(), IntType()) => None
-            case (BoolType(), BoolType()) => None
-            case (StringType(), StringType()) => None
+    private def isSubtype(table: ContractTable, t1: ObsidianType, t2: ObsidianType, isThis: Boolean): Option[Error] = {
```
mcoblenz: As a separate PR, if you were in the mood, you could refactor this. A method called isSubtype should return a Bool, not an Option[Error].

src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -400,6 +486,50 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
...
+                        }
+                    // This will be an error somewhere else
+                    case None => ()
+                }
+            case _ => ()
+        }
+    }
+
+    def substituteOk(table: ContractTable, ast: AST, params: Seq[GenericType], typeArgs: Seq[ObsidianType]): Unit = {
+        // TODO GENERIC: Will have to update this if we do inference for the type args
```
mcoblenz: Change this from TODO GENERIC by filing an issue and referencing it.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
ReedOei: I could. The best way I can think of now is to make a new type of bound (e.g., `PrimitiveBound`), then in the typechecker if it encounters one of those it gives an errors and transforms it to the default `Top` bound. 

My initial thought is that it seems like a lot of complication for something that 1) people will probably almost never write and 2) has a pretty obvious fix.

mcoblenz: Yeah, that's the approach I came up with. It's fine to file an issue on this and leave the code as is.

ReedOei:
> This code causes the compiler to throw an exception:
> `main contract Top {
> 
> }`
> 
> Exception in thread "main" com.helger.jcodemodel.JClassAlreadyExistsException
> at com.helger.jcodemodel.JPackage._class(JPackage.java:178)
> at com.helger.jcodemodel.JPackage._class(JPackage.java:68)
> at com.helger.jcodemodel.IJClassContainer._class(IJClassContainer.java:133)
> at com.helger.jcodemodel.IJClassContainer._class(IJClassContainer.java:116)
> at edu.cmu.cs.obsidian.codegen.CodeGen.$anonfun$translateProgramInPackage$3(CodeGen.scala:131)
> at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:234)
> at scala.collection.immutable.List.foreach(List.scala:389)
> at scala.collection.TraversableLike.map(TraversableLike.scala:234)
> at scala.collection.TraversableLike.map$(TraversableLike.scala:227)
> at scala.collection.immutable.List.map(List.scala:295)
> at edu.cmu.cs.obsidian.codegen.CodeGen.translateProgramInPackage(CodeGen.scala:125)
> at edu.cmu.cs.obsidian.codegen.CodeGen.translateProgram(CodeGen.scala:104)
> at edu.cmu.cs.obsidian.Main$.translateServerASTToJava(Main.scala:124)
> at edu.cmu.cs.obsidian.Main$.compileProgram(Main.scala:341)
> at edu.cmu.cs.obsidian.Main$.main(Main.scala:248)
> at edu.cmu.cs.obsidian.Main.main(Main.scala)

@mcoblenz To be clear, is the suggestion here that this should not crash the compiler, but instead log an error? It seems like this is something that shouldn't be allowed.

mcoblenz:
Right, we shouldn't try to support this; just don't crash. But I think changing the name of Top to Contract might make this easier to understand for people coming from Java.

ReedOei:
@mcoblenz I believe I have addressed all of your comments. The Travis build won't work, because of runtime changes, but I got everything to work on my computer by using a custom repository.

PR #284, opened 2020-02-13
FabCar example and String concatenation

ishantheperson:
This PR overloads `+` in Obsidian to perform both integer addition and string concatenation. This is used to implement the Hyperleder FabCar example in Obsidian. 

The type checker has been modified to accept both int and string as operands to `+`. It produces a new error type `PlusTypeError` in order to provide a more helpful error message. It also produces a new AST node `StringConcat` to easily differentiate integer addition from string concatenation in the code generation stage. 

I've added 1 basic test case, and plan to add more tommorow. 
I also need to add the other version of FabCar into this too

### DIFF
--- a/resources/demos/FabCar/FabCar.obs
+++ b/resources/demos/FabCar/FabCar.obs
@@ -1,43 +1,3 @@
-import "Comparator.obs"
-import "Integer.obs"
-import "Dict.obs"
-
-contract CarComparator implements Comparator[Car] {
-  transaction equals(Car@Unowned a, Car@Unowned b) returns bool {
-    return a.getColor() == b.getColor() 
-        && a.getMake() == b.getMake()
-        && a.getModel() == b.getModel()
-        && a.getOwner() == b.getOwner(); 
-  }
-}
-
-contract CarListPrinter implements Iterator[Integer, Car] {
-  string text;
-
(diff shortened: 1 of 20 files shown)


resources/demos/FabCar/FabCar.obs
```diff
@@ -0,0 +1,146 @@
...
+  transaction getOwner() returns string { return owner; }
+
+  transaction toString() returns string {
+    return "{color='" + color + "',make='" + make + "',model='" + model + "'',owner='" + owner + "'}";
+  }
+}
+
+// I couldn't find anything about maps or arrays
+// in the documentation, but we can make linked lists!
+contract CarList {
```
mcoblenz: Can you use extend the polymorphic linked list in the standard library to do what you want? It doesn't support search, but you can adapt functionality from ERC20/Dict.obs to do that. In particular:
* move ERC20/Dict.obs to Obsidian_Runtime/src/main/java/Runtime/edu/cmu/cs/obsidian/stdlib
* Adapt the technique used there to make a LinkedList of comparable items
* put that new linked list in the standard library

mcoblenz: Can we go ahead and merge this, and you can do the remaining work in a separate PR?

PR #286, opened 2020-04-20
Obsidian LSP

ishantheperson:
This PR is for IDE-like features in the obsidian vscode extension. 

- [x] Show typechecking issues 
- [x] Show parsing issues

(following will require modifications to obsidianc)  
- [ ] Improve error locations by getting the compiler to give both the start and ending location
- [ ] Improve completions by using AST information from obsidianc

### DIFF
--- a/obs-vscode-extension/README.md
+++ b/obs-vscode-extension/README.md
@@ -13,8 +13,19 @@
 
 ## Usage
 
+Compiling: 
 * npm install 
-* npm run compile 
+* npm run compile
+
+To develop the extension/LSP:
+* code -n . (to open the project in vscode) 
+* Go to the "Run" menu on the left 
+* First, start "Launch client"
+* The extension will open in a new VSCode window
+* Now select "Attach to server"
+* You can now trigger breakpoints, etc. 
+
+To install:
 * vsce package
 * code --install-extension obsidian-0.0.1.vsix 
 * Restart VSCode.
--- a/obs-vscode-extension/README.md
+++ b/obs-vscode-extension/README.md
@@ -27,7 +27,7 @@ To develop the extension/LSP:
 
 To install:
 * vsce package
-* code --install-extension obsidian-0.0.1.vsix 
+* code --install-extension obsidian-\*.vsix 
 * Restart VSCode.
 
 ## Citations
--- a/obs-vscode-extension/README.md
+++ b/obs-vscode-extension/README.md
@@ -13,7 +13,8 @@
 
 ## Usage
 
-* tsc
+* npm install 
+* npm run compile 
 * vsce package
 * code --install-extension obsidian-0.0.1.vsix 
 * Restart VSCode.
--- a/obs-vscode-extension/package.json
+++ b/obs-vscode-extension/package.json
@@ -4,7 +4,7 @@
     "publisher": "mcoblenz",
     "repository": "http://www.github.com/mcoblenz/Obsidian",
     "description": "support for Obsidian language.",
-    "version": "0.0.1",
+    "version": "0.1.0",
     "engines": {
         "vscode": "^1.43.2"
     },
--- a/obs-vscode-extension/package.json
+++ b/obs-vscode-extension/package.json
@@ -24,6 +24,7 @@
     ],
     "scripts": {
         "watch": "tsc -b -w",
+        "postinstall": "cd server && npm install",
         "compile": "tsc && cd server && tsc"
     },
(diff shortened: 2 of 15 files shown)


ishantheperson:
image
image
image

obs-vscode-extension/server/src/server.ts
```diff
@@ -0,0 +1,80 @@
+import { createConnection, ProposedFeatures, TextDocuments, InitializeParams, TextDocumentSyncKind, TextDocumentChangeEvent, Position, Diagnostic, Range, DiagnosticSeverity } from "vscode-languageserver";
```
mcoblenz, line 1: Awfully long. Maybe just import everything from this package?

obs-vscode-extension/server/src/server.ts
```diff
@@ -0,0 +1,80 @@
...
+import { promises as fs } from "fs";
+
+const connection = createConnection(ProposedFeatures.all);
+const documents: TextDocuments<TextDocument> = new TextDocuments(TextDocument);
+
+connection.onInitialize((params: InitializeParams) => {
+  return {
+    capabilities: {
+      textDocumentSync: TextDocumentSyncKind.Full,
+      // completionProvider: {
```
mcoblenz, line 14: Is this for future use? You should either remove the commented-out code or extend the comment with an explanation.

obs-vscode-extension/server/src/server.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+});
+
+async function validateTextDocument(change: TextDocumentChangeEvent<TextDocument>) {
+  connection.console.log("Validating document");
+
+  // Create temporary file containing the current set of changes 
+  // TODO: Modify obsidianc to accept source code via stdin instead 
+  // Note: obsidianc crashes if the source file begins with a dot
```
mcoblenz, line 27: Did you file a bug for this?

obs-vscode-extension/server/src/server.ts
```diff
@@ -0,0 +1,80 @@
...
+  const tmpName = __dirname + "/ObsidianLSPTmp.obs"
```
mcoblenz, line 28: Sounds like this might be better: https://www.npmjs.com/package/tmp

obs-vscode-extension/package.json
```diff
@@ -17,12 +17,14 @@
         "@types/mocha": "^2.2.42",
         "@types/node": "^10.14.17",
         "tslint": "^5.12.1",
-        "typescript": "^3.3.1",
-        "vscode": "^1.1.34"
+        "typescript": "^3.3.1"
```
mcoblenz, line 20: I had typescript 3.5.2 and your code refused to compile because you used the ?. operator, which is new in 3.7. You should update the dependencies accordingly (I've upgraded typescript locally, but package.json should be right).

obs-vscode-extension/server/src/server.ts
ishantheperson: Unfortunately there is no such mechanism to do so in typescript/javascript. The alternative is to import the entire namespace which can become very verbose

ishantheperson:
I think that may be related to a change in the build procedure. I have added the instructions to the readme (`npm install && npm run compile`).

obs-vscode-extension/server/src/server.ts
ishantheperson: Done

ishantheperson:
I couldn't seem to reproduce the issue, but I suspect it may have happened due to the version number not being updated in the package. I pushed a commit which does that, does that fix the issue?

obs-vscode-extension/server/src/server.ts
ishantheperson: This turns out to only be a problem if somehow the user is editing two files in parallel. Since that scenario is basically impossible, adding a new dependency seemed overkill

obs-vscode-extension/README.md
```diff
@@ -13,7 +13,19 @@
...
+* code -n . (to open the project in vscode) 
+* Go to the "Run" menu on the left 
+* First, start "Launch client"
+* The extension will open in a new VSCode window
+* Now select "Attach to server"
+* You can now trigger breakpoints, etc. 
+
+To install:
 * vsce package
 * code --install-extension obsidian-0.0.1.vsix
```
mcoblenz, line 30: Need to update this version number to correspond with the version bump.

PR #288, opened 2020-06-11
Updated versioning allows java 13+ to be used when compiling

Tim-Ganger:
These changes fixed version incompatibilities and removed depreciated syntax, imports, and classes.


.idea/checkstyle-idea.xml
```diff
@@ -0,0 +1,16 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcoblenz, line 1: Is this file required, or is it really a separate change?

buildscript/build.gradle
```diff
@@ -37,6 +37,7 @@ dependencies {
     compile 'org.glassfish:javax.json:1.1.0-M1'
     compile 'org.json:json:20171018'
     compile 'com.google.protobuf:protobuf-java:3.11.4'
+    compile ''
```
mcoblenz, line 40: Remove?

.idea/checkstyle-idea.xml
Tim-Ganger: I don't think so, it looks like this was created by intellij and I didn't filter it out.

PR #289, opened 2020-06-28
refactor java code gen into a separate object called CodeGenJava unde…

YinglanChen:
…r package codegen

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenerator.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenerator.scala
@@ -1,0 +1,11 @@
+package edu.cmu.cs.obsidian.codegen
+
+import java.nio.file.Path
+
+import edu.cmu.cs.obsidian.CompilerOptions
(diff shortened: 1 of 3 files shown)


YinglanChen:
From 06/30 meeting, we decided to keep CodeGenJava as object to be consistent with Lexer and Parser.
A trait codeGenerator is added containing one single method gen(). CodeGenJava now extends this trait, so will the future CodeGenYul object.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenerator.scala
```diff
@@ -0,0 +1,11 @@
+package edu.cmu.cs.obsidian.codegen
+
+import java.nio.file.Path
+
+import edu.cmu.cs.obsidian.CompilerOptions
+import edu.cmu.cs.obsidian.parser.SymbolTable
+
+trait CodeGenerator{
```
mcoblenz, line 8: Please add a space before the curly brace.

YinglanChen: Fixed

PR #301, opened 2021-03-26
running evm with ganache-cli in travis ci jobs

ivoysey:
At time of creating this pull request, I believe that the code we generate is indeed getting run on ganache. That said, ganache reports an error when `EmptyContract.sh`:

```
{"id":1,"jsonrpc":"2.0","result":"0x89229df36ed2355c3ae6c0888b1a86d79fa9022fd6a27b31730823da425c5507","error":{"message":"VM Exception while processing transaction: invalid opcode","code":-32000,"data":{"0x89229df36ed2355c3ae6c0888b1a86d79fa9022fd6a27b31730823da425c5507":{"error":"invalid opcode","program_counter":5,"return":"0x"},"stack":"c: VM Exception while processing transaction: invalid opcode\n    at Function.c.fromResults (/usr/local/lib/node_modules/ganache-cli/build/ganache-core.node.cli.js:4:192416)\n    at w.processBlock (/usr/local/lib/node_modules/ganache-cli/build/ganache-core.node.cli.js:42:50915)\n    at processTicksAndRejections (node:internal/process/task_queues:93:5)","name":"c"}}}killing ganache-cli

  Transaction: 0x89229df36ed2355c3ae6c0888b1a86d79fa9022fd6a27b31730823da425c5507
  Contract created: 0x9740e28a2d21178bc95fbde76e2e1aabf1a31871
  Gas usage: 30000000
  Block Number: 1
  Block Time: Fri Mar 26 2021 17:27:35 GMT-0400 (Eastern Daylight Time)
  Runtime Error: invalid opcode
```

This is because the Obsidian compiler generates Yul from which `solc` produces evm that causes this error. In that sense, the testing infrastructure is working by showing us that we have a failing test. I have made a new issue, #309, (which will result in a new branch and PR) to address that problem separately so that we can merge the testing set up itself into master so it doesn't bitrot. 

There is a fair amount of technical debt in this PR, as anticipated. I have documented these things in `#todo` comments in `EmptyContract.sh` as well as in #302 so that we don't lose track of them.


ivoysey:
@mcoblenz this PR is now be ready for you to review. a quick summary:

- The script `EmptyContract.sh` compiles obsidian to Yul then evm, spins up a `ganache-cli` instance, sends that evm to ganache, and runs it. It does some rudimentary checks to see if there are error conditions as a result of this. It does not check anything about the state of the vm after the code gets run (mostly because in the case of the empty contract, I'm not sure what the post condition would be).
- The versions of Java, scala, Ubuntu, gradle, etc. have all been increased to more modern ones.
- This PR includes a little bit of churn of small changes to scala files that I did as a result of these updates, 
- Currently the Obsidian compiler produces Yul from the empty contract that `solc` will compile, but the resultant evm is not valid. We talked about this quickly this morning (30 March) but didn't come to a solid answer. I decided to split that off into a separate issue (#309) and put it on the project board for tomorrow so that I could tidy up this PR. This means that TravisCI *correctly* now observes that the tests fail because the evm doesn't run. I do not know if you want that to be true about master or not, but, if you merge this PR as is, it will be.
- The hyperledger and fabric tests were temporarily disabled to save time on Travis builds when I was debugging that, but I've added them back in. That involved editing some scripts and config files that are quite old; I am not positive I got that right and didn't break anything. 
- There's some churn in some IDE files and other metadata that I think shouldn't bother anyone else but you will know better than I do if that's actually the case.
- There is some technical debt in `EmptyContract.sh`, which I've commented in the file and also documented in #302. I've put that on the project board as being fairly active because I think we'll have to address some of it _en passant_ in the next week or two, but I don't plan to focus on it until it comes up naturally.

ivoysey:
@mcoblenz as of https://github.com/mcoblenz/Obsidian/pull/301/commits/6cd351a1f17b976819d8ca01b771a2ab07a87b05 this PR now gets a non-error return from `ganache-cli` after adding return statements to the output Yul (rather than the bad opcodes response in https://github.com/mcoblenz/Obsidian/issues/309). However, the code generated internally does not match what gets output after going through mustache (see https://github.com/mcoblenz/Obsidian/pull/301/commits/9d5a33cc09c82c417f48d5bbc7f45cec18a9d120). I have made a new issue for that, https://github.com/mcoblenz/Obsidian/issues/310

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -463,8 +463,10 @@ object Parser extends Parsers {
                     case None =>
                         Field(isConst = false, typ, name._1, availableInSet).setLoc(typ)
             }
+            case _ ~ _ ~ _ ~ _ ~ Some(x) ~ _ => throw new RuntimeException("TODO") //failure("available in fields only permitted at the top level of contracts")
```
mcoblenz, line 466: This is a weird comment here — the text in the failure call is not a TODO, it's just not supported. Is that what's going on here? If so, replace TODO with that comment.

mcoblenz (approved):
This looks like great progress! I just had a couple of small changes to make — no need for me to re-review before merging. However, please be sure to SQUASH when you merge. We don't need all of the individual commits in the history.

resources/tests/GanacheTests/EmptyContract.sh
```diff
@@ -0,0 +1,149 @@
...
+# form the JSON object to ask for the list of accounts
+ACCT_DATA=$( jq -ncM \
+                --arg "jn" "2.0" \
+                --arg "mn" "eth_accounts" \
+                --argjson "pn" "[]" \
+                --arg "idn" "1" \
+                '{"jsonrpc":$jn,"method":$mn,"params":$pn,"id":$idn}'
+         )
+
+# we'll keep querying for the accounts until we get a good result, this
```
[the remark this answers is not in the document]
ivoysey: addressed in https://github.com/mcoblenz/Obsidian/pull/301/commits/f5cb70bcd8454f761cd4e9ac917d51a5ab67b4c5 !

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
ivoysey: addressed in https://github.com/mcoblenz/Obsidian/pull/301/commits/ec413aeb0ca0e52c33d8bb8cc506a5ece96a90bc !

Issue #310, opened 2021-03-31
fix mustache file (5 April)


ivoysey:
The code generated in `CodeGenYul.scala` seems to be getting mangled when it gets written out via mustache.  For example, on the empty contract, if you print just before you return from `translateContract`, you get 
```
YulObject(EmptyContract,Code(Block(List(ExpressionStatement(FunctionCall(Identifier(mstore),List(Literal(number,64,int),
Literal(number,128,int)))),
ExpressionStatement(FunctionCall(Identifier(return),List(Literal(number,0,int),
Literal(number,0,int))))))),List(YulObject(EmptyContract_deployed,Code(Block(List(ExpressionStatement(FunctionCall(Identifier(mstore),List(Literal(number,64,int),
Literal(number,128,int)))),
ExpressionStatement(FunctionCall(Identifier(return),List(Literal(number,0,int),
Literal(number,0,int))))))),List(),List())),List())
```
which represents the Yul program

```
object "EmptyContract" {
    code {
            mstore(64,128)
            return(0,0)
    }
    object "EmptyContract_deployed" {
        code {
            mstore(64,128)
            return(0,0)
        }
    }
}
```

but the file that gets written to disk contains

```
object "EmptyContract" {
    code {
            mstore(64,128)
            return(0,0)
    }
    object "EmptyContract_deployed" {
        code {
            return(0,0)
        }
    }
}
```

(which is missing the `mstore` in the `_deployed` version).

ivoysey:
originally in this commit, for reference: https://github.com/mcoblenz/Obsidian/commit/9d5a33cc09c82c417f48d5bbc7f45cec18a9d120

ivoysey:
Another instance of bad output: the obsidian program 
```
main contract IntCons{
        int x;
        transaction intconst() {
            x = 4;
            return;
        }
}
```
produces
```
List(ExpressionStatement(FunctionCall(Identifier(mstore),List(Literal(number,64,int), Literal(number,128,int)))), FunctionDefinition(intconst,List(),List(),Block(List(ExpressionStatement(FunctionCall(Identifier(sstore),List(Literal(number,0,int), Literal(number,4,int)))), Leave()))), ExpressionStatement(FunctionCall(Identifier(return),List(Literal(number,0,int), Literal(number,0,int)))))
```
for its `statement_seq_runtime` but the Yul that gets output has the `return(0,0)` first:
```

object "IntCons"{
            mstore(64,128)
            return(0,0)        
    }
[...]

Issue #303, opened 2021-03-29
Integer literals (3/31)


ivoysey:
- [x] Make a new branch
- [x] Write a test program, which is just a function that returns an integer literal (return 42)
- [x] Determine appropriate YUL type (pick something simple, e.g. int32)
- [x] Assert on integer literals that are out of bounds for that type
- [x] Translate int literals to appropriate YUL code
- [x] Translate return statement to Yul.
- [x] add a test to ganache tests and see that it runs in travis

ivoysey:
https://docs.soliditylang.org/en/v0.8.3/yul.html?highlight=yul#simple-example

ivoysey:
per meeting with @mcoblenz on 1 April: compiling `return e` requires storing values in memory; so this task is just going to be compiling `return` with no arguments.

PR #311, opened 2021-04-01
adding support for integer literals

ivoysey:
This PR addresses #303. It also touches on #310 because I found that the Yul emitted didn't compile without changing the mustache file.


mcoblenz (approved):
LGTM, but note the request in the other PR to de-duplicate the test scripts, which applies here too.

Issue #308, opened 2021-03-29
branches (if then else) (5 April)


ivoysey:
- [x] New branch
- [x] Write test case: if (true) return 1 else return 0
- [x] Add switch to Yul AST if needed
- [x] Translate if to switch (because Yul lacks if/else)
- [x] Submit PR

ivoysey:
See #315

PR #315, opened 2021-04-05
adding translation for if-then-else, some code hoisting for clarity

ivoysey:
This implements translation of if-then-else to switch per #308, I think, but really shows that the problem in #310 is important to fix right away. Specifically, when translating the test case
```
main contract IfThenElse{
        int x;
        transaction ifthenelse() {
            if (true) {
	       x = 1;
	    } else {
	       x = 0;
	    }
            return;
        }
}
```
we actually produce the expansion
```
List(Switch(Literal(boolean,true,bool),List(Case(Literal(boolean,true,bool),Block(List(ExpressionStatement(FunctionCall(Identifier(sstore),List(Literal(number,0,int), Literal(number,1,int))))))), Case(Literal(boolean,false,bool),Block(List(ExpressionStatement(FunctionCall(Identifier(sstore),List(Literal(number,0,int), Literal(number,0,int))))))))))
```
but the written file is
```
object "IfThenElse" {
    code {
            mstore(64,128)
            return(0,0)
    }
    object "IfThenElse_deployed" {
        code {
            return(0,0)
                let selector := shr(224, calldataload(0))
                switch selector
                    case 0x70a08231 {
                    }
function ifthenelse(){
}
        }
    }
}
```
This not only drops the `mstore` command, but places the `return` above the code not below and omits the body of the function we produced entirely.

### DIFF
--- a/resources/tests/GanacheTests/IfThenElse.obs
+++ b/resources/tests/GanacheTests/IfThenElse.obs
@@ -1,4 +1,4 @@
-main contract IfThenElse{
+main contract IfThenElse {
         int x;
         transaction ifthenelse() {
             if (true) {
--- a/resources/tests/GanacheTests/IfThenElse.obs
+++ b/resources/tests/GanacheTests/IfThenElse.obs
@@ -1,0 +1,11 @@
+main contract IfThenElse{
+        int x;
+        transaction ifthenelse() {
+            if (true) {
+	       x = 1;
+	    } else {
+	       x = 0;
+	    }
+            return;
+        }
+}
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
@@ -257,18 +257,15 @@ object CodeGenYul extends CodeGenerator {
                 }
             case IfThenElse(scrutinee,pos,neg) =>
                 val scrutinee_yul : Seq[YulStatement] = translateExpr(scrutinee)
-                if(scrutinee_yul.length > 1){
+                if (scrutinee_yul.length > 1){
                     assert(false,"boolean expression in conditional translates to a sequence of expressons")
                     Seq()
                 }
                 scrutinee_yul.apply(0) match {
                     case ExpressionStatement(sye) =>
-                        //print("scrutinee explanded to: " + sye.toString())
                         val pos_yul = pos.flatMap(translateStatement)
                         val neg_yul = neg.flatMap(translateStatement)
-                        val ret = Seq(Switch(sye,Seq(Case(true_lit , Block(pos_yul)), Case(false_lit, Block(neg_yul)))))
-                        print ("expanded to: " + ret.toString())
-                        ret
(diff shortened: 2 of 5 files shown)


ivoysey:
@mcoblenz I marked this as a draft to indicate that it is definitely not ready for master, and we should chat tomorrow, but I still moved the cards into the "merge and review" column because that seemed like the right thing too.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -250,11 +255,27 @@ object CodeGenYul extends CodeGenerator {
                         assert(false, "TODO")
                         Seq()
                 }
-
-            case _ =>
-                assert(false, "TODO")
+            case IfThenElse(scrutinee,pos,neg) =>
+                val scrutinee_yul : Seq[YulStatement] = translateExpr(scrutinee)
+                if(scrutinee_yul.length > 1){
```
mcoblenz, line 260: space after if

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -250,11 +255,27 @@ object CodeGenYul extends CodeGenerator {
...
+                    assert(false,"boolean expression in conditional translates to a sequence of expressons")
+                    Seq()
+                }
+                scrutinee_yul.apply(0) match {
+                    case ExpressionStatement(sye) =>
+                        //print("scrutinee explanded to: " + sye.toString())
+                        val pos_yul = pos.flatMap(translateStatement)
+                        val neg_yul = neg.flatMap(translateStatement)
+                        val ret = Seq(Switch(sye,Seq(Case(true_lit , Block(pos_yul)), Case(false_lit, Block(neg_yul)))))
+                        print ("expanded to: " + ret.toString())
```
mcoblenz, line 270: comment this out

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -250,11 +255,27 @@ object CodeGenYul extends CodeGenerator {
...
-                assert(false, "TODO")
+            case IfThenElse(scrutinee,pos,neg) =>
+                val scrutinee_yul : Seq[YulStatement] = translateExpr(scrutinee)
+                if(scrutinee_yul.length > 1){
+                    assert(false,"boolean expression in conditional translates to a sequence of expressons")
+                    Seq()
+                }
+                scrutinee_yul.apply(0) match {
+                    case ExpressionStatement(sye) =>
+                        //print("scrutinee explanded to: " + sye.toString())
```
mcoblenz, line 266: delete this one

mcoblenz (approved):

resources/tests/GanacheTests/IfThenElse.obs
```diff
@@ -0,0 +1,11 @@
+main contract IfThenElse{
```
mcoblenz, line 1: space after IfThenElse

ivoysey:
per discussion with @mcoblenz this morning, i'm going to merge this into master even though it produces weird output because we want to fix those issues in a separate branch. we believe that the code backing the generation of yul for if-then-else is fundamentally correct.

PR #323, opened 2021-04-19
temp vars refactor

ivoysey:
this threads an index for new temporary variables through translation of expressions to yul, with the invariant that the result of the expression is stored in the incoming variable as the last action in the resultant sequence of yul statements. this makes it possible to use the result of the recursive calls in translate expression.

this also includes some automatic whitespace and scala style fixing. currently the travis builds do not work because they aren't fetching sbt but i believe that's a passing problem, not on our end.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK">
+  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="true" project-jdk-name="1.8" project-jdk-type="JavaSDK">
     <output url="file://$PROJECT_DIR$/out" />
   </component>
 </project>
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
@@ -23,6 +23,15 @@ object CodeGenYul extends CodeGenerator {
     var stateEnumMapping: Map[String, Int] = Map() // map from state name to an enum value
     var stateEnumCounter: Int = 0 // counter indicating the next value to assign since we don't know the total num of states
 
+    // we generate new temporary variables with a little bit of global state; i am making the
+    // implicit assumption that nothing except nextTemp will modify the contents of tempCnt, even
+    // though that is not enforced statically.
+    var tempCnt: Int = 0
+    def nextTemp(): Identifier = {
+        tempCnt = tempCnt + 1
+        Identifier(name = s"_tmp_${tempCnt.toString}") //todo: better naming convention?
+    }
+
     def gen(filename: String, srcDir: Path, outputPath: Path, protoDir: Path,
             options: CompilerOptions, checkedTable: SymbolTable, transformedTable: SymbolTable): Boolean = {
         // extract ast and find main contract
@@ -40,7 +49,7 @@ object CodeGenYul extends CodeGenerator {
                 Paths.get(mainName)
         }
         // translate from obsidian AST to yul AST
-        val translated_obj = translateProgram(ast)
+        val translated_obj = translateProgram(nextTemp(), ast)
         // generate yul string from yul AST
         val s = translated_obj.yulString()
         // write string to output file
@@ -52,12 +61,12 @@ object CodeGenYul extends CodeGenerator {
         true
     }
 
-    def translateProgram(program: Program): YulObject = {
+    def translateProgram(retvar: Identifier, program: Program): YulObject = {
         // translate main contract, or fail if none is found or only a java contract is present
         val main_contract_ast: YulObject =
             findMainContract(program) match {
                 case Some(p) => p match {
-                    case c@ObsidianContractImpl(_, _, _, _, _, _, _, _) => translateContract(c)
+                    case c@ObsidianContractImpl(_, _, _, _, _, _, _, _) => translateContract(nextTemp(), c)
                     case JavaFFIContractImpl(_, _, _, _, _) =>
                         throw new RuntimeException("Java contract not supported in yul translation")
                 }
@@ -75,7 +84,7 @@ object CodeGenYul extends CodeGenerator {
                         // note: interfaces are not translated;
                         // TODO detect an extra contract named "Contract", skip that as a temporary fix
                         if (c.name != ContractType.topContractName) {
-                            new_subObjects = main_contract_ast.subObjects :+ translateContract(obsContract)
+                            new_subObjects = main_contract_ast.subObjects :+ translateContract(nextTemp(), obsContract)
                         }
                     }
                 case _: JavaFFIContractImpl =>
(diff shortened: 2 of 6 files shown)


.idea/misc.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project version="4">
-  <component name="ProjectRootManager" version="2" languageLevel="JDK_1_8" default="false" project-jdk-name="1.8" project-jdk-type="JavaSDK">
```
mcoblenz, line 3: What is this change for?

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
```diff
@@ -129,20 +142,24 @@ case class Block (statements: Seq[YulStatement]) extends YulStatement {
...
-case class YulObject (name: String, code: Code, subObjects: Seq[YulObject], data: Seq[Data]) extends YulAST {
+case class Code(block: Block) extends YulAST
+
+case class Data(name: String, hex: Seq[HexLiteral], str: Seq[StringLiteral]) extends YulAST
+
+case class HexLiteral(content: String) extends YulAST
+
+case class StringLiteral(content: String) extends YulAST
+
+case class YulObject(name: String, code: Code, subs: Seq[YulObject], data: Seq[Data]) extends YulAST {
```
mcoblenz, line 153: Please expand "subs."

mcoblenz: (it reads like "substitutions" to me, and we're just not that short on space here).

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
```diff
@@ -48,74 +52,83 @@ case class Literal (kind: LiteralKind.LiteralKind, value: String, vtype: String)
...
+case class Switch(expression: Expression, cases: Seq[Case]) extends YulStatement {
     override def toString: String = {
-        s"switch ${expression.toString}" + "\n" + (if(cases.isEmpty) brace("") else cases.map(c => c.toString).mkString("\n")) + "\n"
+        s"switch ${expression.toString}" + "\n" +
+            (if (cases.isEmpty) brace("") else cases.map(c => c.toString).mkString("\n")) +
+            "\n"
     }
 }
 
-case class ForLoop (pre: Block, condition: Expression, post: Block, body: Block) extends YulStatement {
```
mcoblenz, line 98: why did you rename `condition`?

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -115,8 +125,8 @@ object CodeGenYul extends CodeGenerator {
     }
 
     // return statements that go to deploy object, and statements that go to runtime object
-    def translateDeclaration(declaration: Declaration): (Seq[YulStatement], Seq[YulStatement]) = {
```
mcoblenz, line 118: Why the rename?

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -171,7 +181,7 @@ object CodeGenYul extends CodeGenerator {
                 new_name, // TODO rename transaction name (by adding prefix/suffix) iev: this seems to be done already
                 constructor.args.map(v => TypedName(v.varName, mapObsTypeToABI(v.typIn.toString))),
                 Seq(), //todo/iev: why is this always empty?
-                Block(constructor.body.flatMap(translateStatement))))
+                Block(constructor.body.flatMap((s: Statement) => translateStatement(s))))) //todo iev temp vars: likely a hack to work around something wrong
```
mcoblenz, line 184: What is this todo for?

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -186,7 +196,7 @@ object CodeGenYul extends CodeGenerator {
             transaction.name, // TODO rename transaction name (by adding prefix/suffix)
             transaction.args.map(v => TypedName(v.varName, mapObsTypeToABI(v.typIn.toString))),
             ret,
-            Block(transaction.body.flatMap(translateStatement))))
+            Block(transaction.body.flatMap((s: Statement) => translateStatement(s))))) //todo iev temp vars: likely a hack to work around something wrong
```
mcoblenz, line 199: What is this todo for?

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -209,28 +219,19 @@ object CodeGenYul extends CodeGenerator {
...
-                        Seq(edu.cmu.cs.obsidian.codegen.Switch(sye, Seq(Case(true_lit, Block(pos_yul)), Case(false_lit, Block(neg_yul)))))
-                    case e =>
-                        assert(assertion = false, "if statement built on non-expression: " + e.toString)
-                        Seq()
-                }
-            case e: Expression => translateExpr(e)
+            case IfThenElse(scrutinee, pos, neg) =>
+                val id = nextTemp()
+                val scrutinee_yul: Seq[YulStatement] = translateExpr(id, scrutinee)
+                val pos_yul: Seq[YulStatement] = pos.flatMap(s => translateStatement(s)) // todo iev be careful here this might be wrong
```
mcoblenz, line 230: what's this todo for?

mcoblenz (approved):
Lots of little naming comments but no need to re-review.

.idea/misc.xml
ivoysey: i'm not sure, i think it came with an update to the IDE that i pulled down this morning. i guess i'll leave but, but it's not something i claim to understand.

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
ivoysey: oh, yeah, good catch!

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
ivoysey: no good reason; i've reverted it.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: no good reason; i've reverted it.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: in general i'm not sure when i'm using `flatMap` because it's the right thing versus hiding a bug in how i'm thinking about statements, so i have some todos just as bookmarks to help me find places that i don't quite trust.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: as above, flatMap is a yellow flag

PR #324, opened 2021-04-20
Improve the GitHub look of the last section

oleks:
Instead of looking looking something like this:
obs1

It now looks something like this:
obs2


mcoblenz (approved):
Thanks!

PR #325, opened 2021-04-20
Update Travis link to use .com domain

oleks:
The Travis repository has already been migrated to the .com domain,
all that remains is to update this link.

In general, this change as announced by Travis in May 2018:
https://docs.travis-ci.com/user/migrate/open-source-on-travis-ci-com/


mcoblenz (approved):
Thanks for the PR!

Issue #318, opened 2021-04-09
integer comparison


ivoysey:
Implement integer comparison, found here https://github.com/mcoblenz/Obsidian/blob/functioncalls/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala#L307-L350. This is fairly low hanging fruit, and a good thing to do just to make steady progress.

ivoysey:
the guts of this are implemented in https://github.com/mcoblenz/Obsidian/tree/primops as of 21 april. the test case is broken because we need to implement at least `return e` (#327) or (#328) to be able to run those tests and both of those require memory management.

Issue #313, opened 2021-04-02
check output of running code on ganache


ivoysey:
right now there is no particular check against the output from running ganache, just that it runs and doesn't make an error. check the output. this is a complex enough thing that i'm migrating it to its own issue, off of the general technical debt issue #302. i think we'll need to address this pretty soon.

as of 24 may, there are 4 tests that don't pass that i want to pass before merging this:

- [ ] PrimOpsNEq
- [ ] PrimOpsNeg
- [ ] PrimOpsNot
- [ ] PrimOpsSubNeg

ivoysey:
This will reveal a bunch of bugs, I bet. These may include:

- [ ] implement renaming functions (why do we need this, exactly? check the ABI)
- [ ] emitting the empty constructor in the runtime part of the contract

ivoysey:
some useful-looking end points:

- https://eth.wiki/json-rpc/API#eth_getstorageat
- https://eth.wiki/json-rpc/API#eth_sendtransaction
- https://eth.wiki/json-rpc/API#eth_getbalance
- https://eth.wiki/json-rpc/API#eth_gettransactioncount

send transaction Returns
```
DATA, 32 Bytes - the transaction hash, or the zero hash if the transaction
is not yet available.

Use eth_getTransactionReceipt to get the contract address, after the
transaction was mined, when you created a contract.
```
so what can you actually do with transaction hashes? here's one thing:

https://eth.wiki/json-rpc/API#eth_gettransactionbyhash

use a to-account to put the answer somewhere? or look it up in storage?

ivoysey:
- [ ] figure out why primops doesn't run through solc
```
Exception while assembling: /solidity/libyul/backends/evm/EVMObjectCompiler.cpp(66): Throw in function void solidity::yul::EVMObjectCompiler::run(solidity::yul::Object&, bool)
Dynamic exception type: boost::wrapexcept<solidity::yul::StackTooDeepError>
std::exception::what: Variable x is 1 slot(s) too deep inside the stack.
[solidity::util::tag_comment*] = Variable x is 1 slot(s) too deep inside the stack.
```

ivoysey:
After reading through the json-rpc, my rough design for a first pass at this is as follows. Extend the json object in the `GanacheTests` directory for each test case with:
- `data`, which is a Yul fragment of what we want to store to check xxx
- `enc`, which is Yul and encodes the data above into a u256 int
- `dec`, which is BASH to decode
[...]

PR #329, opened 2021-04-23
translation to primative operations

ivoysey:
this addresses #317 and #318. it won't be testable until we do at least one of #328 or #327; the test case is written but i deliberately didn't include the json file in the PR so that it doesn't run.

### DIFF
--- a/resources/tests/GanacheTests/PrimOps.obs
+++ b/resources/tests/GanacheTests/PrimOps.obs
@@ -1,9 +1,10 @@
 main contract PrimOps{
         int x;
         bool y;
+
         transaction primops() {
-            y = true and false;
-            y = true or false;
+            //y = true and false; //these don't parse, or i have the syntax wrong
+            //y = true or false;
 
             x = 4 + 4;
             x = 4 - 4;
@@ -19,7 +20,7 @@ main contract PrimOps{
             y = 3 <= 3;
             y = 3 != 3;
 
-            y = not y;
+            //y = not y; // as above
             x = - x;
 
             return;
--- a/resources/tests/GanacheTests/PrimOps.obs
+++ b/resources/tests/GanacheTests/PrimOps.obs
@@ -1,0 +1,7 @@
+main contract PrimOps{
+        int x;
+        transaction primops() {
+            x = 4;
+            return;
+        }
+}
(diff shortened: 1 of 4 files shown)


resources/tests/GanacheTests/PrimOps.obs
```diff
@@ -0,0 +1,28 @@
+main contract PrimOps{
+        int x;
+        bool y;
+
+        transaction primops() {
+            //y = true and false; //these don't parse, or i have the syntax wrong
```
mcoblenz, line 6: Obsidian currently uses C-style names for these (&&, ||, !).

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -214,12 +215,13 @@ object CodeGenYul extends CodeGenerator {
...
-                            case l =>
-                                assert(assertion = false, "TODO: unimplemented translate assignment case: " + l.toString)
+                            case StringLiteral(_) => LiteralKind.string
+                            case _ =>
+                                assert(assertion = false, s"unimplemented assignment case ${assignTo.toString}")
                                 LiteralKind.number
                         }
                         Seq(ExpressionStatement(FunctionCall(Identifier("sstore"),
-                            Seq(ilit(tempSymbolTable(x)), Literal(kind, e.toString, kind.toString)))))
+                            Seq(ilit(tempSymbolTable(x)), Literal(kind, e.toString, kind.toString))))) //todo: this is likely wrong for strings
```
mcoblenz, line 224: If you think it's wrong, assert rather than silently doing the wrong thing.

mcoblenz (approved):
Looks good.

resources/tests/GanacheTests/PrimOps.obs
ivoysey: ah! ok! i read the parser and thought i had it right, but thanks for the fix.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: i didn't want to make an assertion here because the code *runs* just fine with string constants, solc is happy to pretty print the ouput, but i think it won't actually work once we can check the output. so this is more of a note to myself for the future (e.g. #313) but i want to be able to check the output now.

Issue #333, opened 2021-05-21
clean up boilerplate generation


ivoysey:

ivoysey:
https://docs.soliditylang.org/en/latest/internals/variable_cleanup.html a lot of the things we currently omit are these sorts of hygiene conditions; that's fine for now but will not be fine always

PR #336, opened 2021-05-24
update test script to check output

### LINKED ISSUE #313: check output of running code on ganache

ivoysey:
this will close #313 once it's merged and is ready for review!


bin/run_obs_yul_all.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
mcoblenz, line 1: Please add a comment explaining what this file does.

bin/skipped_tests.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
mcoblenz, line 1: Please add a comment explaining what this file does.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -283,11 +316,13 @@ object CodeGenYul extends CodeGenerator {
                     s.flatMap(s => {
                         val id_s: Identifier = nextTemp()
                         decl_0exp(id_s) +: translateStatement(s, Some(id_s.name), contractName, checkedTable)
+                        //todo: this also does not assign afterwards; likely the same bug as fixed in IfThenElse
```
mcoblenz, line 319: This sounds serious and potentially hard to debug  later. Is there an issue open for this? If not, please file one and mark it high priority.

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -283,11 +316,13 @@ object CodeGenYul extends CodeGenerator {
...
                     })
 
                 decl_0exp(id_scrutinee) +:
                     scrutinee_yul :+
                     edu.cmu.cs.obsidian.codegen.If(id_scrutinee, Block(s_yul))
+            // todo: also no assignment here
```
mcoblenz, line 325: Is there an issue for this?

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
```diff
@@ -276,8 +278,8 @@ case class YulObject(name: String, code: Code, subobjects: Seq[YulObject], data:
                 codegen.ExpressionStatement(apply("abi_decode_tuple", intlit(4), apply("calldatasize"))),
                 //    fun_retrieve_24()
                 call_f_and_maybe_assign,
-                //    let memPos := allocate_memory(0)
-                decl_1exp(mp_id, apply("allocate_memory", intlit(0))),
+                //    let memPos := allocate_unbounded()
```
mcoblenz, line 281: What does allocate_unbounded do?

mcoblenz (approved):

bin/run_obs_yul_all.sh
ivoysey: got it!

bin/skipped_tests.sh
ivoysey: got it!

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: that's right. i've made an issue for it here and will get to it sooner rather than later. https://github.com/mcoblenz/Obsidian/issues/339

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: if it's wrong, it's a manifestation of the same bug as above, so i've noted it in https://github.com/mcoblenz/Obsidian/issues/339

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
ivoysey: it's a less complicated version of allocation that the more recent build of solc outputs; i shifted to it while trying to debug the `0x` return from Ganache that turned out to be something else. i think either way is OK for now, but figuring out how to do this right is part of #333

Issue #335, opened 2021-05-24
testing improvements


ivoysey:
- [x] allow for tests with input
- [x] ~allow for tests that have return type other than int~
- [x] ~test `fib(5)`~
- [x] refactor test script so that you have one json file that describes all the tests to run instead of one file per test, now that there are a lot of tests.
- [x] check the output from curl commands in an exhaustive way
- [x] deal with broken json files in a graceful way?
- [x] output a summary of how many tests passed and failed and which ones failed on Travis
- [x] output a warning if there are `.obs` files without a `.json` file (meaning test that are there but not being run)
- [x] look for ways to make the tests take less time to run (shared work getting repeated, etc.)

this subsumes #302 

as of 2 June, these tests are defined but do not have JSON so are not getting run (because they don't pass):

- [x] `PrimOpsNeg` (#337 and also #338)
- [x] `PrimOpsSubNeg` (#338)
- [x] `MultiLineIfRet` (#341)
- [ ] ~`SimpleStructure` (#307)~

additionally, 
- [x] the `Return` test is written to avoid a bug I documented in #334

ivoysey:
@mcoblenz do you have any thoughts for low hanging fruit for how to speed up the test run? like caching we could be doing? i know that `sbt` gets run once per JSON file in the `GanacheTests` directory, and it takes some time to notice that it doesn't need to do any work to rebuild Obsidian each time, for instance. it's not clear to me that `sbt` really needs to be run more than once for the whole test suite, but maybe that's aggressive.

the full run of a CI pass is taking something like 20 mins and it's starting to be a burden; i can run the tests locally first, but want to see them pass in Travis before merging a branch into master, for instance.

ivoysey:
this commit https://github.com/mcoblenz/Obsidian/commit/ba1a07b720a174a9c0c1daa13773d18df25390ae seems to make the tests only get run by sbt in one branch through the travis jobs but not both.

ivoysey:
i'm going to close this issue once #408 gets merged. it's too big and most of it's been done; i've broken out what hasn't into smaller issues and it's all pretty low priority, but this big monolithic thing has served its purpose.

PR #343, opened 2021-06-07
reconfiguring travis for slightly shorter build times

ivoysey:
this breaks the ganache tests out into a separate job with Travis matrices, so it runs in parallel. this also removes a bit of a hack we needed to work around a known bug with Travis and sbt that has now been fixed, and the ganache tests now use the jar'ed up version of Obsidian instead of making a call to sbt on every test, which saves a few minutes of run time.

this is part of #335 but does not close it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,8 +31,8 @@ install:
   - sbt ++$TRAVIS_SCALA_VERSION test # this actually builds Obsidian
 
 env:
-  - TEST_SUITE=tests
-  - TEST_SUITE=ganache_tests
+  - TEST_SUITE=tests.sh
+  - TEST_SUITE=ganache_tests.sh
 
 script: "bash travis_specific/$TEST_SUITE"
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,15 +18,15 @@ before_install:
   - ./travis_specific/install_ganache.sh
 
 install:
-  - |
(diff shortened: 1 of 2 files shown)


ivoysey:
@mcoblenz when you review this, you should check over the log output for the non-ganache job pretty carefully and make sure it looks right to you. i did a pass and i didn't see anything that stood out as busted, but it is not something i'm very familiar with.

.travis.yml
```diff
@@ -18,21 +18,14 @@ before_install:
...
-    export SBT_OPTS="-Dfile.encoding=UTF-8"
-    curl -L --silent "https://github.com/sbt/sbt/releases/download/v$SBT_LAUNCHER/sbt-$SBT_LAUNCHER.tgz" > $HOME/sbt.tgz
-    tar zxf $HOME/sbt.tgz -C $HOME
-    sudo rm /usr/local/bin/sbt
-    sudo ln -s $HOME/sbt/bin/sbt /usr/local/bin/sbt
-
-script:
   - gradle publish -b Obsidian_Runtime/build.gradle
-  - sbt ++$TRAVIS_SCALA_VERSION test # this actually builds Obsidian
-  - bash travis_specific/tests.sh
```
mcoblenz, line 34: Let's remove YulTests from tests.sh, since it's redundant with your new tests.

mcoblenz (approved):

PR #346, opened 2021-06-10
testing updates from structures branch

ivoysey:
the origin branch here is mostly for implementing structures, but some testing improvements happened there too, addressing #335 in part. specifically: ganache tests and fabric tests now happen in parallel, and the ganache testing job doesn't install the dependencies it doesn't need. also the fabric tests should only happen on PR builds into master.

i'm opening this PR because these changes will save us a bunch of Travis credits, but also because it's the only way to test if the non-ganache job actually gets spun up on PRs getting merged into master.


travis_specific/ganache_tests.sh
```diff
@@ -1,5 +1,12 @@
 #!/bin/bash
 
+# this should only happen when this gets run on Travis; for now it's enough to check that the uname
+# is Linux
+if [[ $(uname) == "Linux" ]]
```
mcoblenz, line 5: Check CI == true instead? https://docs.travis-ci.com/user/environment-variables/#default-Environment-Variables

travis_specific/install_ganache.sh
```diff
@@ -1,9 +1,7 @@
 #!/bin/bash
 
 npm install -g npm
-npm install
```
mcoblenz, line 4: npm appears to already be installed (https://docs.travis-ci.com/user/languages/javascript-with-nodejs/); do we still need the line above?

travis_specific/tests.sh
```diff
@@ -1,5 +1,13 @@
 #!/bin/bash
```
mcoblenz, line 1: Now would be a great time to rename this file to fabric_tests.sh.

mcoblenz (approved):

travis_specific/ganache_tests.sh
ivoysey: ah, great!

travis_specific/tests.sh
ivoysey: yes.

PR #349, opened 2021-07-06
Moved ContractStateDetector into its own file and modified testing script to use pytest

tmoux:


corpus_studies/solidity_state_study/slither_plugins/run_tests.py
```diff
@@ -1,41 +1,27 @@
...
-def test_state():
-    failed = []
-    files = [f for f in glob.glob("tests/state/*.sol")]
-    for f in files:
-        print("testing %s..." % f)
-        slither = Slither(f)
-        if not any(ContractStateDetector(c).is_stateful_contract() for c in slither.contracts):
-            failed.append(f)
-    return failed
+# To run tests, run python3 -m pytest run_tests.py
```
mcoblenz, line 7: I'm a python newbie — can one also invoke pytest directly (https://docs.pytest.org/en/6.2.x/)?

mcoblenz (approved):

corpus_studies/solidity_state_study/slither_plugins/run_tests.py
tmoux: Yes, I just fixed the imports so you can do that.

PR #364, opened 2021-08-17
simple structure types 

ivoysey:
This PR addresses #355, #356, #357, and #358: reading and writing fields to storage, child contract subobjects, invocations without arguments that return at most 32 bytes.

This means that we can compile, run, and succeesfully test code for these two example programs:
```
contract IntContainer{
    int x;

    transaction set1() {
        x = 5;
    }

    transaction set2() {
	x = 10;
    }

    transaction get() returns int{
        return x;
    }
}

main contract SetGetNoArgsNoConstruct{
    transaction main() returns int{
        IntContainer ic = new IntContainer();
        ic.set1();
	ic.set2();
        return (ic.get());
    }
}
```
```
contract IntContainer{
    int x;

    transaction set() {
        x = 5;
    }

    transaction get() returns int{
        return x;
    }
}

main contract SetGetNoArgsNoConstructNoInit{
    transaction main() returns int{
        IntContainer ic = new IntContainer();
        ic.set();
        return ic.get();
    }
}
```
There are several major weaknesses and simplifying assumptions being made, but this is a solid step. In particular, the code at https://github.com/mcoblenz/Obsidian/pull/364/files#diff-0d1fb9c1be91df00b3ba70b970f3a7c92b9d6d867f2f30adfc602e40867a61ccR497 is just plain wrong, and we can figure out what to do in review.


ivoysey:
@mcoblenz it looks like the fabric tests failed on this? https://app.travis-ci.com/github/mcoblenz/Obsidian/jobs/532150913
```
Creating network "net_byfn" with the default driver

Creating volume "net_orderer.example.com" with default driver

Creating volume "net_peer0.org1.example.com" with default driver

Creating volume "net_peer1.org1.example.com" with default driver

Creating volume "net_peer0.org2.example.com" with default driver

Creating volume "net_peer1.org2.example.com" with default driver

Pulling orderer.example.com (hyperledger/fabric-orderer:latest)...

ERROR: manifest for hyperledger/fabric-orderer:latest not found: manifest unknown: manifest unknown

ERROR !!!! Unable to start network

./invoke.sh -q getY

Expected '1' from './invoke.sh -q getY' but got 'Fabric network not found. Did you remember to run up.sh?'

The command "travis_specific/$TEST_SUITE" exited with 1.
before_cache.1

0.00s$ rm -fv $HOME/.ivy2/.sbt.ivy.lock
before_cache.2

0.00s$ find $HOME/.ivy2/cache -name "ivydata-*.properties" -print -delete
before_cache.3

0.00s$ find $HOME/.sbt        -name "*.lock"               -print -delete
cache.2

store build cache
``` 

is the error but it doesn't mean a lot to me.

mcoblenz:
It looks like the thing that went wrong is independent of your code. You can proceed.

.idea/runConfigurations.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
mcoblenz, line 1: What is this file for, and why do we want Android?

Obsidian_Runtime/src/main/yul_templates/object.mustache
```diff
@@ -82,16 +87,23 @@ object "{{creationObject}}" {
                 mstore(pos, cleanup_t_uint256(value))
             }
 
+            function revert_forward_1() {
```
mcoblenz, line 90: What does this do? A comment would help.

Obsidian_Runtime/src/main/yul_templates/object.mustache
```diff
@@ -82,16 +87,23 @@ object "{{creationObject}}" {
...
                 cleaned := value
             }
+
             {{#runtimeFunctions}}
                 {{code}}
             {{/runtimeFunctions}}
         }
-        {{runtimeSubObjects}}
+
+        {{childcontracts}}
```
mcoblenz, line 107: CamelCase?

src/main/scala/edu/cmu/cs/obsidian/codegen/Util.scala
```diff
@@ -135,23 +157,24 @@ object Util {
...
       *
       * @param id the identifier to be declared
-      * @param e the expression to assign the identifier to as an initial value
+      * @param e  the expression to assign the identifier to as an initial value
       * @return the Yul expression for the declaration
       */
     def decl_1exp(id: Identifier, e: Expression): VariableDeclaration = decl_nexp(Seq(id), e)
 
-    def mapObsTypeToABI(ntype: String): String = {
+    def mapObsTypeToABI(ntype: String): (String, Int) = {
```
mcoblenz, line 165: What does the Int represent? Add a comment. Ideally this would return an object, since then the fields could have names.

mcoblenz: I see it's the size from reading the code, but that is not apparent from the name.

src/main/scala/edu/cmu/cs/obsidian/codegen/yulAST.scala
```diff
@@ -198,7 +198,7 @@ case class HexLiteral(content: String) extends YulAST
 
 case class StringLiteral(content: String) extends YulAST
 
-case class YulObject(name: String, code: Code, subobjects: Seq[YulObject], data: Seq[Data]) extends YulAST {
+case class YulObject(name: String, code: Code, runtimesubobj: Seq[YulObject], childcontracts: Seq[YulObject], data: Seq[Data]) extends YulAST {
```
mcoblenz, line 201: CamelCase.

mcoblenz (approved):
Be sure to squash when you merge!

.idea/runConfigurations.xml
ivoysey: I don't know the answer to either question; I think it got made when I had to do a hard reset and update to 2021.2.

Obsidian_Runtime/src/main/yul_templates/object.mustache
ivoysey: It reverts but also copies out the return data from memory. I think it's to aid debugging; I found it in the generated yul from solc and didn't want to mess with it because I didn't know how important it was. I'll add a comment.

src/main/scala/edu/cmu/cs/obsidian/codegen/Util.scala
ivoysey: I've added a comment and updated the name. I'm going to leave it as returning a tuple rather than a case class for now because I think this function is a little bit of a hack and temporary, but I'm not sure how it'll end up getting rewritten.

PR #368, opened 2021-09-01
flatten obsidian subcontracts before translation to yul

ivoysey:


ivoysey:
The idea I started fleshing out in this branch turns out to be a dead end (at least for now); a good experiment but not worth keeping. I'm going to close the PR to indicate that. I will pull some of the earlier commits into another PR later.

Issue #373, opened 2021-09-13
refactor the translation to yul to be type directed rather than syntax directed


ivoysey:
update the typechecker so that it annotates every node of the AST with the type of that node. at the beginning of type checking, these should all start out as `None` and incrementally get updated to `Some(foo)` as they get discovered; this amounts to memoizing the derivation tree for type checking.

per MC: add a field, get compiler warnings that the constructors don't set that field, then change the constructors to get another argument to set that field, then get more warnings, etc.

once this works and is tested, change the code in the flat_alloc branch to use it; this should clear up a lot of hacks there.

PR #376, opened 2021-09-14
memoize type info

ivoysey:
this addresses #373. i added a field to the expression nodes of the AST to store their type as its computed so that it's then available to yul code generation. this meant making changes also to the parser and test cases, as well as a few other things, to reflect the different signatures for the relevant expressions.

right now this passes the typechecker tests; i am not at all confident that it's perfectly correct, but I think i need to use it to see what's broken and then fix those errors once they come up.

i have left comments where I was least sure about what I was doing. hopefully that will make the errors easier to find, or maybe some of them can get caught in review before they get merged.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
@@ -10,19 +10,23 @@ trait HasLocation {
     var loc: Position = NoPosition
 
     def setLoc(t: Token): this.type = {
-        loc = t.pos; this
+        loc = t.pos;
+        this
     }
 
     def setLoc(other: HasLocation): this.type = {
-        loc = other.loc; this
+        loc = other.loc;
+        this
     }
 
     def setLoc(id: (String, Position)): this.type = {
-        loc = id._2; this
+        loc = id._2;
+        this
     }
 
     def setLoc(pos: Position): this.type = {
-        loc = pos; this
+        loc = pos;
+        this
     }
 }
 
@@ -36,7 +40,8 @@ sealed abstract class Statement() extends AST {
 /* All expressions are statements. We relegate the pruning of expressions
  * that don't have effects to a later analysis */
 sealed abstract class Expression() extends Statement {
-    val obstype : Option[ObsidianType]
+    val obstype: Option[ObsidianType]
+
     override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): Expression
 }
 
@@ -85,12 +90,14 @@ sealed abstract class InvokableDeclaration() extends Declaration {
 // Expressions not containing other expressions
 sealed abstract class AtomicExpression extends Expression {
     val obstype: Option[ObsidianType] = None
+
     override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): AtomicExpression = this
 }
 
 sealed abstract class UnaryExpression(make: Expression => UnaryExpression,
                                       e: Expression) extends Expression {
     val obstype: Option[ObsidianType] = None
+
     override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): UnaryExpression =
         make(e.substitute(genericParams, actualParams)).setLoc(this)
 }
@@ -99,6 +106,7 @@ sealed abstract class BinaryExpression(make: (Expression, Expression) => BinaryE
                                        e1: Expression,
                                        e2: Expression) extends Expression {
     val obstype: Option[ObsidianType] = None
+
     override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): BinaryExpression =
         make(e1.substitute(genericParams, actualParams), e2.substitute(genericParams, actualParams))
             .setLoc(this)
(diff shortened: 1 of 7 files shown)


src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
```diff
@@ -64,95 +89,182 @@ sealed abstract class InvokableDeclaration() extends Declaration {
...
+case class LessThanOrEquals(e1: Expression, e2: Expression) extends BinaryExpression(LessThanOrEquals, e1, e2) {
+    override val obstype: Option[ObsidianType] = Some(BoolType())
+}
+
+case class NotEquals(e1: Expression, e2: Expression) extends BinaryExpression(NotEquals, e1, e2) {
+    override val obstype: Option[ObsidianType] = Some(BoolType())
+}
+
 case class Dereference(e: Expression, f: String) extends UnaryExpression(Dereference(_, f), e) {
+    override val obstype: Option[ObsidianType] = e.obstype // todo
```
mcoblenz, line 218: Why todo? Since obstype is immutable, I think this is safe.

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
```diff
@@ -64,95 +89,182 @@ sealed abstract class InvokableDeclaration() extends Declaration {
...
     override def substitute(genericParams: Seq[GenericType], actualParams: Seq[ObsidianType]): Construction =
         Construction(contractType.substitute(genericParams, actualParams),
-            args.map(_.substitute(genericParams, actualParams)), isFFIInvocation)
+            args.map(_.substitute(genericParams, actualParams)), isFFIInvocation, typ)
             .setLoc(this)
 }
-case class Disown(e: Expression) extends UnaryExpression(Disown, e)
-case class StateInitializer(stateName: Identifier, fieldName: Identifier) extends Expression {
+
+case class Disown(e: Expression) extends UnaryExpression(Disown, e) // todo there's something happening that i don't understand
```
mcoblenz, line 263: Namely?

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -212,23 +211,23 @@ object Parser extends Parsers {
...
+                case f ~ _ ~ e => (ReferenceIdentifier(f._1, None).setLoc(f), e)
             }
             LParenT() ~ repsep(oneUpdate, CommaT()) ~ RParenT() ^^ {
                 case _ ~ updates ~ _ => updates
             }
         }
 
         val parseTransition = RightArrowT() ~ parseId ~
-                              opt(parseUpdate) ~! SemicolonT() ^^ {
+            opt(parseUpdate) ~! SemicolonT() ^^ {
```
mcoblenz, line 222: This whitespace change isn't great.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
```diff
@@ -265,7 +263,7 @@ object Parser extends Parsers {
         }
 
         val parseTryCatch = TryT() ~! LBraceT() ~! parseBody ~! RBraceT() ~!
-                            CatchT() ~! LBraceT() ~! parseBody <~ RBraceT() ^^ {
+            CatchT() ~! LBraceT() ~! parseBody <~ RBraceT() ^^ {
```
mcoblenz, line 266: This whitespace change isn't great.

src/main/scala/edu/cmu/cs/obsidian/util/IdentityAstTransformer.scala
```diff
@@ -148,14 +148,14 @@ class IdentityAstTransformer {
             case s: StringLiteral => (s, List())
             case t: TrueLiteral => (TrueLiteral().setLoc(t), List())
             case f: FalseLiteral => (FalseLiteral().setLoc(f), List())
-            case t: This => (This().setLoc(t), List())
+            case t: This => (This(None).setLoc(t), List()) // todo: no idea if this is right
```
mcoblenz, line 151: For the identity transformer, just use whatever  type is already there (not None).

mcoblenz (changes requested):
Just a few minor comments. Please be sure to address all the TODOs you wrote before merging.

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
ivoysey: OK! I marked it as a todo because I was kind of guessing, but if it "smells right" to you then I'll go with it.

src/main/scala/edu/cmu/cs/obsidian/parser/AST.scala
ivoysey: I spent some time trying to add a `obstype` field to this case class, overriding the `None` I think it would inherit from `UnaryExpression` with the obstype of `e`, but I was struggling to get the syntax right to express what I meant. `UnaryExpression` is parametric (that's probably not the scala-specific word) in first parameter, which is then always recursive here, so I don't quite know how to change it --- or if I should.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
ivoysey: I agree. It's what IntelliJ did when I cleaned up the file and I think in line with what `scalafix` would do, but I'm not in love with it. I think a lot of linters don't quite know what to do with clever infixing.

src/main/scala/edu/cmu/cs/obsidian/parser/Parser.scala
ivoysey: This is the same story as the other place you pointed out.

src/main/scala/edu/cmu/cs/obsidian/util/IdentityAstTransformer.scala
ivoysey: got it!

Issue #379, opened 2021-10-28
fix failing typechecker tests in the flat_alloc branch


ivoysey:
somewhere in the checker the type annotations are not getting added, so 5 of the typechecker tests are failing.

PR #380, opened 2021-10-29
fixing type checker errors

ivoysey:
this PR fixes the remaining type checking errors. some of the tests were defined incorrectly, but i had to make some changes in the type checker as well, so that's worth looking at a little more carefully.

this will address #379

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
@@ -549,6 +549,21 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
         }
     }
 
+    /**
+      * TODO: document the invariants of this function. at several call sites, we assume that the
+      *   annotation in the expression returned matches the type returned (and therefore to not call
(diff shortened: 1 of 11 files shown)


src/main/scala/edu/cmu/cs/obsidian/typecheck/Checker.scala
```diff
@@ -1789,7 +1790,7 @@ class Checker(globalTable: SymbolTable, verbose: Boolean = false) {
                         val fieldAST = newStateTable.lookupField(f)
                         if (fieldAST.isDefined) {
                             val (t, contextPrime2, ePrime) = inferAndCheckExpr(decl, contextPrime, e, consumptionModeForType(fieldAST.get.typ))
-                            newUpdates = newUpdates :+ (ReferenceIdentifier(f, Some(t)), ePrime)
+                            newUpdates = newUpdates :+ (ReferenceIdentifier(f, Some(t)), ePrime) // todo iev: does ePrime here need its obstype field updated?
```
mcoblenz, line 1793: If inferAndCheckExpr is supposed to do it, then no. You should decide and update it if needed.

mcoblenz (approved):
Looks good. You should resolve those TODOs and then merge.

Issue #372, opened 2021-09-13
make a PR where the set-get example works, writing everything to 0x80 in memory


ivoysey:
this is the so-called "constant allocator". this will be in the flat_alloc branch to start with; right now that code makes the flat Yul object but still stores everything in storage.

Issue #371, opened 2021-09-13
make a PR where the set-get example works, writing things sequentially to memory but never deallocating anything


ivoysey:
this is the so called "sequential allocator". do this after #372

Issue #370, opened 2021-09-13
have the contents of the dispatch table respect abstraction and visibility


ivoysey:
right now, in flat_alloc, the dispatch table contains all the functions that get flattened into the main contract as well as all the functions of the main contract. only the public ones should be there; i shouldn't expose code to the world in the process of flattening it.

a hack would be to check the function names as i process them and see if they  `m/(.+)___(.+)/`. a somewhat better idea would be to have another data structure in YulObject that includes functions that get created by flattening methods and process them differently. adding type information per #373 may actually also address this, if that includes privacy information as well; then we know what functions ought to be world-visible, both from the main contract and from child contracts.

ivoysey:
solc outputs something like this for a setter that takes one integer argument,
```
                case 0x60fe47b1
                {
                    // set(uint256)                                                                                              

                    if callvalue() { revert_error_ca66f745a3ce8ff40e2ccaf1ad45db7774001b90d25810abd9040049be7bf4bb() }
                    let param_0 :=  abi_decode_tuple_t_uint256(4, calldatasize())
                    fun_set_13(param_0)
                    let memPos := allocate_unbounded()
                    let memEnd := abi_encode_tuple__to__fromStack(memPos  )
                    return(memPos, sub(memEnd, memPos))
```
which pulls the argument from the stack. it's not clear to me how to do this when the first argument is the added implicit `this` argument that allows us to allocate into memory. from the outside there may not even be an instance of the object created, and in general actually won't be, so what would you even try to pass there?

ivoysey:
this issue is somewhat entwined with #385

Issue #361, opened 2021-08-17
understand why the diff command for missing json files produces different output on travis and locally; are all the tests getting run?


ivoysey:

ivoysey:
https://github.com/mcoblenz/Obsidian/pull/384/commits/66d4838e55ae2dfcd2b6455b77a4308ecf45cf1e this one, too, but i had a typo in the commit so it didn't associate

PR #384, opened 2021-11-03
compute offsets, store fields in memory

ivoysey:
this addresses #371 and #372. These tests pass:

- https://github.com/mcoblenz/Obsidian/blob/flat_offsets/resources/tests/GanacheTests/SG.obs
- https://github.com/mcoblenz/Obsidian/blob/flat_offsets/resources/tests/GanacheTests/SetGetNoArgsNoConstruct.obs
- https://github.com/mcoblenz/Obsidian/blob/flat_offsets/resources/tests/GanacheTests/SetGetNoArgsNoConstructNoInit.obs

These do not:

- https://github.com/mcoblenz/Obsidian/blob/flat_offsets/resources/tests/GanacheTests/SetGet.obs fails because the entry for the set function that takes an argument in the dispatch table has a free variable in it; this is documented in https://github.com/mcoblenz/Obsidian/issues/370 but I'm not quite sure how to fix it.
- https://github.com/mcoblenz/Obsidian/blob/flat_offsets/resources/tests/GanacheTests/SetGetNoArgs.obs fails because I haven't implemented constructors or constructors that take arguments (https://github.com/mcoblenz/Obsidian/issues/359)

this also addresses https://github.com/mcoblenz/Obsidian/issues/361

### DIFF
--- a/resources/tests/GanacheTests/SetGetNoArgs.obs
+++ b/resources/tests/GanacheTests/SetGetNoArgs.obs
@@ -1,7 +1,7 @@
 contract IntContainer{
     int x;
 
-    IntContainer@Owned() {
+    IntContainer() {
         x = 0;
     }
 
--- a/bin/skipped_tests.sh
+++ b/bin/skipped_tests.sh
@@ -3,4 +3,4 @@
 # when run from the top level of the repo, this prints out the names of the obsidian files
 # in the ganache tests that do not also have a json file paired with them -- that means that
 # they won't be run by travis. this also appears in the output of the travis script at the very top.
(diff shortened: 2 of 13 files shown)


resources/tests/GanacheTests/SetGetNoArgs.obs
```diff
@@ -1,7 +1,7 @@
 contract IntContainer{
     int x;
 
-    IntContainer@Owned() {
+    IntContainer() {
```
mcoblenz, line 4: Isn't this supposed to be specified?

mcoblenz (approved):

PR #388, opened 2021-11-10
allocate fields with pointers, rewrite yulobject and mustache interaction to reflect current object layout

### LINKED ISSUE #393: add a `this` argument to transactions from the main contract

ivoysey:
This closes #393, #392, #391, and #389. It partially addresses #370; I think that everything in the main contract will end up in the dispatch table now, even if it's private.

One worry is that the code in `YulObject` assumes that the `Seq[YulStatement]` passed contains only function declaration objects -- this is made explicit by a use of `.asInstaceOf`. That happens to be true right now, but won't be forever I think. I'm OK with this bit of technical debt for now, but, @mcoblenz, you may not be. 

Once this is approved and merged into `flat_alloc`, I'll do a merge into master as well because I think this is a stand alone feature update as well as being on the way to allocation.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
@@ -153,6 +153,8 @@ object CodeGenYul extends CodeGenerator {
             }
         }
 
+        // todo: i think that i can remove the inMain argument form translateStatement and translateExpression
+
         // form the body of the transaction by translating each statement found
         val body: Seq[YulStatement] = transaction.body.flatMap((s: Statement) => translateStatement(s, id, contractName, checkedTable, inMain))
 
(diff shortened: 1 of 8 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -418,6 +317,53 @@ object CodeGenYul extends CodeGenerator {
             assign1(retvar, apply("or", apply(s, e1id, e2id), apply("eq", e1id, e2id)))
     }
 
+    //todo document this; it's a helper to share some code repeated between the two Invocation cases
```
mcoblenz, line 320: Should fill in documentation.

mcoblenz (approved):

src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
ivoysey: ah, missed this one! thanks.

PR #404, opened 2021-12-01
emit tracers for each contract

ivoysey:
this PR includes code that emits tracers for each contract found. they are stubs for now, they just walk the structure and log things as they go. a later PR will update this to do the task of updating a reference count.


ivoysey:
i'm going to merge this without review into my shadow master so that i can use it in other branches.

PR #410, opened 2021-12-17
updates to testing harness and emitted code for stashing values

ivoysey:
this is an incremental PR that does not fully implement writing logs at pointer writes. that said, it does the following:
* updates the CI script to ganache 7
* implements a new top level commandline flag `--stash` to have the Yul codegen emit a stash function that writes ints to sequential blocks in storage starting from a magic address
* a new test case that shows that stashing works
* updates to the test script that allow tests to check what gets stashed
* updates to the test script that show that we can detect logs when they get written (but doesn't check them)
* metadata updates for the most recent version of intellij
* updates to the testing script that both deploy and invoke the transactions on the block chain rather than only running them locally to the node (which is how storage writes and logs actually get triggered)


ivoysey:
i'm closing this pr without merging it. after discussion with michael, i'll pull just the logging out into its own PR without the stashing work around code. this branch can be left hanging so that we have easy access to it if we need it, but it's not going to get included into my shadow master.

Issue #397, opened 2021-11-18
get a nested intcontainer example working


ivoysey:
i think the back end of this will already work. the code should look something like this, but i'm getting the ownership annotations wrong
```
contract IntContainer{
    int x;

    IntContainer@Owned(int init) {
        x = init;
    }

    transaction set(int value) {
        x = value;
    }

    transaction get() returns int{
        return x;
    }
}

contract IntContainerWrapper{
    IntContainer@Owned ic;

    IntContainerWrapper@Owned(IntContainer@Owned init) {
        ic = init;
    }

    transaction set(IntContainerWrapper@Owned this, int value) {
        ic.set(value);
    }
    transaction get(IntContainerWrapper@Owned this) returns int{
        return (ic.get());
    }
}

main contract SetGetNested{
    transaction main() returns int{
        IntContainer ic = new IntContainer(30);
        IntContainerWrapper iw = new IntContainerWrapper(ic);
        int x = iw.get();
        iw.set(5);
        return (x + ic.get());
    }
}
```

PR #411, opened 2021-12-17
fix broken nested intcontinater test

ivoysey:


ivoysey:
This will close #397 by adding two test cases that show that the problem doesn't seem to appear any more. It also pulls in the rewrite of the test script to use pyweb3 from another branch, because it's sufficiently better that I don't want to be without it.

mcoblenz (approved):

PR #413, opened 2022-02-11
call tracers after every constructor call

ivoysey:
- we now emit a tracer for every contract in the program, not just the ones that happen to be used in fields in the main contract
- we call the tracer for a contract after it's constructed
- the tracers copy the fields to storage, for now at the same address that they appear in memory
- the tracers emit logs as they do this, but the values emitted are not that meaningful

upcoming changes not included in this PR:
- write meaningful values to the logs
- use those logged values to verify that fields are being written to storage
- write to addresses in storage that are above some thresh hold value, so we can tell which addresses are in memory and which are in storage.

### DIFF
--- a/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
+++ b/src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
@@ -135,11 +135,10 @@ object CodeGenYul extends CodeGenerator {
                                 others = others ++ writeTracers(ct, contractType.contractName)
                             case _ => Seq()
                         }
-                        case _: PrimitiveType => {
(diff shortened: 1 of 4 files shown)


src/main/scala/edu/cmu/cs/obsidian/codegen/CodeGenYul.scala
```diff
@@ -544,12 +557,20 @@ object CodeGenYul extends CodeGenerator {
                         case _ => false
                     }
 
+                // the constructor(s) for the main contract are not prefixed with the name of the contract.
+                val invoke_name =
```
mcoblenz, line 561: I think this would be better refactored into transactionNameMapping.

mcoblenz (approved):

PR #415, opened 2022-02-15
updating master from shadow master

ivoysey:


ivoysey:
i'm merging this into master without waiting for approval because it's coming from shadow. this also doesn't have CI builds because we're out of travis credits just now and waiting for their help folks to replenish us, but i've run the test suite locally and i believe it is OK.

Issue #427, opened 2022-02-25
should the constructor for the main contract appear in the dispatch table?


ivoysey:
more generally, look at which transactions appear in dispatch and do a little house cleaning. see also #370, #385.

i think i can send over a sequence of transactions paired with a boolean for if they ought to appear in the table or not, from codegen to the yul ast writer.

PR #435, opened 2022-04-01
first cut at having some control over what appears in the dispatch table

ivoysey:
this will address #427 #370


ivoysey:
https://github.com/mcoblenz/Obsidian/pull/437 does this in a nicer way.

